PR #7, opened 2015-06-18
travis: enable for all gradle project

proppy:
/cc @ggfan @hak


proppy:
This depends on #8 to be more portable.

proppy:
LGTM self-merging.

PR #8, opened 2015-06-19
builder: add gradle builder test

proppy:
This is a junit testcase, dynamically generating parameterized tests that build all android gradle project of this repository using the gradle tooling API.

/cc @rschiu


rschiu:
LGTM

PR #19, opened 2015-06-30
MoreTeapot: fix android lint warning

proppy:
/cc @hak @ggfan @claywilkinson

This remove lint warning from MoreTeaport sample.
- bump `minsdk` from `11` to `19` since the code has some 19 only API calls.
- update icon to default android studio mipmap
- switch to RTL attributes for text placement
- move text to resources


claywilkinson:
LGTM

hak:
For setSystemUiVisibility() (API-19), the code is checking API level and falling back to other pass.
So if this is the only reason to bump up the API level, we don't have to do that.

proppy:
@hak PTAL

hak:
LGTM, too.

PR #21, opened 2015-07-01
builder: proritize dirty projects

proppy:
- launch `clean` and `lint` first
- build modified projects first


proppy:
@spearce said it was mostly not horrible, self-merging.

Issue #34, opened 2015-07-15
Sample hello-jni won't compile


gengjiawen:
```
UNEXPECTED TOP-LEVEL EXCEPTION:
com.android.dx.cf.iface.ParseException: bad class file magic (cafebabe) or version (0034.0000)
        at com.android.dx.cf.direct.DirectClassFile.parse0(DirectClassFile.java:472)
        at com.android.dx.cf.direct.DirectClassFile.parse(DirectClassFile.java:406)
        at com.android.dx.cf.direct.DirectClassFile.parseToInterfacesIfNecessary(DirectClassFile.java:388)
        at com.android.dx.cf.direct.DirectClassFile.getMagic(DirectClassFile.java:251)
        at com.android.dx.command.dexer.Main.processClass(Main.java:704)
        at com.android.dx.command.dexer.Main.processFileBytes(Main.java:673)
        at com.android.dx.command.dexer.Main.access$300(Main.java:83)
        at com.android.dx.command.dexer.Main$1.processFileBytes(Main.java:602)
        at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:170)
        at com.android.dx.cf.direct.ClassPathOpener.processDirectory(ClassPathOpener.java:229)
        at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:158)
        at com.android.dx.cf.direct.ClassPathOpener.processDirectory(ClassPathOpener.java:229)
        at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:158)
        at com.android.dx.cf.direct.ClassPathOpener.processDirectory(ClassPathOpener.java:229)
        at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:158)
        at com.android.dx.cf.direct.ClassPathOpener.processDirectory(ClassPathOpener.java:229)
        at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:158)
        at com.android.dx.cf.direct.ClassPathOpener.process(ClassPathOpener.java:144)
        at com.android.dx.command.dexer.Main.processOne(Main.java:632)
        at com.android.dx.command.dexer.Main.processAllFiles(Main.java:510)
        at com.android.dx.command.dexer.Main.runMonoDex(Main.java:280)
        at com.android.dx.command.dexer.Main.run(Main.java:246)
[...]

PR #43, opened 2015-07-26
Fixed build error for hello-jni project.

shomeser:
Applied changes from issue #34

### DIFF
--- a/hello-jni/.gitignore
+++ b/hello-jni/.gitignore
@@ -1,0 +1,13 @@
+.gradle
+.idea
+build
+local.properties
+*.iml
+*.tmp
+*.log
+*.bak
+*.class
+*~
+
+# files from device
+traces.txt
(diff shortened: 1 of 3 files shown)


shomeser:
I signed it!

proppy:
Sorry for the late followup, I was in vacation.

Note that another way to "fix" this is to point to another version of java in "Project Structure".

I'm ok having a workaround in our samples, but a real bug should also be filed thru https://code.google.com/p/android/issues/entry?template=Tools%20bug%20report about the wrong version of java being selected by default.

We should probably file the issue first and reference it as a comment in build.gradle.

In the meantime can you rebase the change, and split the .gitignore and api bump in separate PR? so we can apply and review them separately.

Thanks in advance.

proppy:
This is fixed on all samples with #91, thanks for raising this issue.

Issue #20, opened 2015-06-30
all samples should include configuration and instructions for JNI debugging


proppy:
Using:
```
    android.buildTypes {
        debug {
            isJniDebuggable = true
        }
    }
```

kneth:
I am trying to debug some native code using Android Studio 1.3 (preview) with the new NDK support. My `app/build.gradle` has a section like

    android.buildTypes {
        release {
            isMinifyEnabled = false
            proguardFiles += file('proguard-rules.txt')
        }

        debug {
            isJniDebuggable = true
            isDebuggable = true
        }
    }

For gdb I constantly get

    Syntax error: extraneous input 'n' expecting {<EOF>, ','} in "=tsv-created,name="trace_timestamp",initial="0"\n" line 1 position 48
    com.jetbrains.cidr.execution.debugger.backend.gdb.GDBResponse$ResponseParseException: Syntax  error: extraneous input 'n' expecting {<EOF>, ','} in "=tsv-created,name="trace_timestamp",initial="0"\n"   line 1 position 48
  	    at com.jetbrains.cidr.execution.debugger.backend.gdb.GDBResponse$Parser.parse(GDBResponse.java:84)
	    at com.jetbrains.cidr.execution.debugger.backend.gdb.GDBResponse.parse(GDBResponse.java:18)

For lldb it cannot get a handshake.

Do I miss something in my `build.grade`?

proppy:
@kneth can you file a separate issue on https://code.google.com/p/android/issues/entry?template=Tools%20bug%20report so that we can track the RC 1 debugging issue separatly?

This issue is more about updating the the gradle files for the sample to add debug configuration.

proppy:
@kneth (Note that lldb worked for me, albeith slow):
```
DEVICE SHELL COMMAND: cat /data/local/tmp/lldb-server | run-as com.android.gl2jni sh -c 'cat > /data/data/com.android.gl2jni/lldb/bin/lldb-server; chmod 700 /data/data/com.android.gl2jni/lldb/bin/lldb-server'
DEVICE SHELL COMMAND: rm /data/local/tmp/lldb-server
DEVICE SHELL COMMAND: cat /data/local/tmp/start_lldb_server.sh | run-as com.android.gl2jni sh -c 'cat > /data/data/com.android.gl2jni/lldb/bin/start_lldb_server.sh; chmod 700 /data/data/com.android.gl2jni/lldb/bin/start_lldb_server.sh'
[...]

PR #46, opened 2015-07-30
Update plugin to gradle-experimental-0.2.0

rschiu:
gradle-experimental-0.2.0 was just released.  This updates the sample to use the newest version and make some DSL changes.

### DIFF
--- a/bitmap-plasma/app/build.gradle
+++ b/bitmap-plasma/app/build.gradle
@@ -19,7 +19,7 @@ model {
 
     android.buildTypes {
         release {
-            isMinifyEnabled = false
+            minifyEnabled = false
             proguardFiles += file('proguard-rules.txt')
 
             println(proguardFiles)
--- a/MoreTeapots/build.gradle
+++ b/MoreTeapots/build.gradle
@@ -4,7 +4,7 @@ buildscript {
        jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle-experimental:0.1.0'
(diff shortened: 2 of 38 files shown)


bitmap-plasma/app/build.gradle
```diff
@@ -19,7 +19,7 @@ model {
 
     android.buildTypes {
         release {
-            isMinifyEnabled = false
+            minifyEnabled = false
```
proppy: was the `is` prefix also removed from `isJniDebuggable`?

proppy:
LGTM

bitmap-plasma/app/build.gradle
rschiu: isJniDebuggable is now a NDK buildType options and renamed to debuggable.  You can specify it with something like this:
android.buildType {
    release {
        ndk {
            debuggable = true
        }
    }
}

proppy: it seems that this is optional though, even without it I'm able to hit a breakpoint.

rschiu: The debug build is now debuggable by default.  I don't think it would work if you are trying to debug the release build.

proppy: good to know, so no need for us to do #20 anymore

proppy: @rschiu it would be nice to update http://tools.android.com/tech-docs/new-build-system/gradle-experimental with a release + debuggeable example.

PR #56, opened 2015-08-07
targetSdkVersion should be 22

yaraki:
targetSdkVersion should be 22.


proppy:
@yaraki can you change it on the other samples as well?

And possibly update https://github.com/googlesamples/android-ndk/tree/master/builder or https://github.com/googlesamples/android-ndk/blob/master/.travis.yml to enforce it?

proppy:
can you add something like this in `.travis.yml` for both `compileSdkVersion` and `targetSdkVersion` to ensure they remains the same when we update them in the future.
```
test $(echo $(grep -h compileSdkVersion */app/build.gradle  | uniq  | wc -l)) == "6"
```

proppy:
rebase?

PR #70, opened 2015-08-25
travis: always display test summary

proppy:
@ggfan this should improve the test output.
@yaraki can you review, since it modify the change you did in #56


yaraki:
LGTM

PR #62, opened 2015-08-21
Update packaging.yaml

tjohns:
- Set status to PUBLISHED
- Remove "Getting Started" category, since this category is overused.
  (If needed, make a "NDK Getting Started" category?)

PR #73, opened 2015-08-28
hello-jni/packaging: update categories

martynbm:


proppy:
The category looks already fixed with #62, @tjohns asked us not to squat the already overloaded 'Getting Started' category.

PR #80, opened 2015-09-01
add appveyor configuration

proppy:
for windows build test.

PR #81, opened 2015-09-01
builder: remove pandoc deps

proppy:
So that it's easier to add windows build (#80).


claywilkinson:
Can it be conditional on the OS?  That way we can still get nice reports in the non-windows builds?

proppy:
I don't think we want the lint report for all platform, since lint is mostly based on static analysis: I would expect it to be the same everywhere.

I moved the pandoc call to `.travis.yml` that way it will display something on linux.

claywilkinson:
SGTM

PR #86, opened 2015-09-02
fix most lint warnings

proppy:
Related #82

PR #87, opened 2015-09-02
Removing unused layout

moxi:
Complementing #86 for san-angeles demo.

 Initially I extracted the String to avoid the warning but then I noticed the layout is not used so there's no need to have the file at all.


moxi:
I signed it!

proppy:
Thanks a lot for helping with this :)

PR #91, opened 2015-09-03
force JavaVersion.VERSION_1_7 for all samples

proppy:
With:
```
for f in */app/build.gradle; do sed -e $'s/^.*android.ndk/\
    compileOptions.with {\
        sourceCompatibility=JavaVersion.VERSION_1_7\
        targetCompatibility=JavaVersion.VERSION_1_7\
    }\
\
&/' $f; done
```
Closes #43

### DIFF
--- a/MoreTeapots/app/build.gradle
+++ b/MoreTeapots/app/build.gradle
@@ -11,6 +11,12 @@ model {
             targetSdkVersion.apiLevel = 23
         }
     }
+
+    compileOptions.with {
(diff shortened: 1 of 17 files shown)


tjohns:
I approve of this change :+1:.

I wonder if we should do this with the rest of the framework samples...

proppy:
@ggfan changed the SDL syntax as you suggested.

ggfan:
LGTM, thanks!

Issue #79, opened 2015-09-01
add new sample with unit tests


proppy:
Add new sample for showing how to run unittest for C++ and Java and if it works well add unit tests to every samples.

Related to #77.

moxi:
I added a sample and I'm doing some research to add some Tests for the C/C++ code.

ggfan:
bug bankruptcy.

Issue #77, opened 2015-09-01
Build Variants: Test Artifact Unit Tests doesn't work


moxi:
The audio-echo has testCompile defined on its build.gradle, I'm trying to add UnitTests to my project and it doesn't compile:
![exampleunittest](https://cloud.githubusercontent.com/assets/437653/9605849/ed9edc72-5074-11e5-88b0-74e21e8cff04.png)
.
I didn't change the code so you can reproduce it using the project found here:
https://github.com/googlesamples/android-ndk/tree/master/audio-echo
 

What I've tried in order to have unit test available and its not working:
* Upgrade to the latest: com.android.tools.build:gradle-experimental:0.3.0-alpha2
* Create a new Module (plug in java), and add the Android project as dependency and it doesn't work either.
* Compiling using gradlew test.

Are you using something different to have the UnitTests up an running?

ggfan:
it is buildtwith gradle experiment 0.1.0, or 0.2.0  I think; I did not use the test project, probably should delete it before I merge in [ it came with Android Studio when I "new project"] .  I will see how it goes with 0.3.0-alpha2

proppy:
I agree, we should probably delete the test project in the meantime if it's not being used.

proppy:
> probably should delete it before I merge in

FYI, it's already merged into master and so will show up in android studio at soon as we whitelist the project.

proppy:
I went ahead and deleted the unused test and filed a new issue to add sample to show proper unittesting (#79).

@moxi feel free to comment/contribute on #79 if you already have idea on what a good sample for this could look like.

moxi:
Yes, I spent some hours this early morning exploring all the sample projects to add a UnitTest example that will make sense. I'll comment on #79.

PR #92, opened 2015-09-04
Unit test sample

moxi:
cc: @proppy 
Adding a sample that has Android InstrumentationTests and Java UnitTests running correctly to address #79 

My sample is based on the Hello World sample, if simple to show the basics of the configuration and how to test them. The UnitTest part it was a little bit tricky because I had to isolate the Java code, and that means to create stubs of the native methods and not to load the static block that loads the JNI library.

This is the first part, I'm planning to add tests to the C code, but I don't have a lot of knowledge so I'm doing my research.

P.S: As a side note, you will notice this is using gradle-2.6-all.zip and :gradle-experimental:0.3.0-alpha2, I tried with previous versions but it wasn't compiling.

runningunittests

runninginstrumentationtests


moxi:
I signed it!

I did signed it in the past in #87

proppy:
@moxi thanks a lot for your contribution!

I will to get a proper reviewer for the Java side.

A few questions:
1. About C++ testing: what do you think of using https://github.com/google/googletest
1. About Instrumentation test: what do you think of testing the Activity (and check the label) in addition to the JNI bindings?
1. About unit tests: do you think it would be worth testing the JNI call without mocking too?

JoseAlcerreca:
It might be better to add a simple Espresso test and make the unit tests in C++. The Espresso test tests the full stack and the unit test gives you fast feedback. 

Also, rename setUp() with something that explains what the method is doing. "setUp" is JUnit3.

moxi:
@proppy 
a. Yes, that's what I'm planning to use.
b. I was thinking the same but with the addition of Espresso, like @JoseAlcerreca pointed out.
c. Good idea, I wanted to show that you can stub/mock external dependencies in more pure UnitTests. I will need to find out how to let gradle know the path of the NDK library.

@JoseAlcerreca , Yes, I'll rename setUp eventhough is JUnit3 it still makes sense to setUp your tests, I'll come out with a better name.

@JaseAlcerreca, I want to show how to use simple UnitTests without having the need to run an emulator/device, that's why I added the UnitTest example.

Next steps: 
- Add Espresso and make a simple example of how to test the Activity.
- Add C/C++ UnitTests

Thanks, for all the feedback, I'll continue contributing to this example.

JoseAlcerreca:
I'm not sure if the unit test sample is actually testing something. It might be testing that the mocking framework works :) I'll remove it for now and work on the other tests. The Espresso one should be really easy using the instrumentation test you already have.

moxi:
Yes, I'll remove it for now and think about a better example on how not just show the set up of the environment.
To be honest that was my goal after #77, because it was failing on one project where I'm testing my model classes and trying this Android Studio/NDK. I came to see the samples here to see how to configure the Unit Test environment using the NDK and found #77.

moxi:
I have some updates for this sample, 

I removed the UnitTest that was showing the setup as suggested, I added Espresso and a Test using Espresso.

What it's pending is to add C++ Unit Tests. I wonder what do you guys think about merging this (when everything makes sense) and create a new pull request for the C++ UnitTests.

proppy:
Ping, sorry for the late followup. I updated the other sample to 0.4.0. Let me know if your sample is working too.

ip:
@proppy I use Google Test in my project and it is great. But there are two problems:
 - Main problem: there is no access to JNI (so is to assets and other API), because the way of running is somewhat hackish - we compile our program as executable (!), upload it with .so dependencies into /tmp directory, grant it executable permissions and run it from here. In that way, our tests are bare Linux console executable which doesn't know anything about Android and it's APIs.
 - Google Mock is not yet ported, as far as I know.

I posted to their forum but no reaction. At the moment I just exclude those tests that doesn't work on Android with `#if .. #endif`.

I think it should be absolutely possible to properly adapt Google Test to Android to support interoperation with Java side - probably we need to modify test runner part to start new Activity instead of `fork`-ing the process like it is now.

proppy:
@ip it makes a lot of sense, I'd love to explore that with #142.

proppy:
@stewartmiles can also comment on how they run unittest on android for https://github.com/google/fplbase and https://github.com/google/zooshi

stewartmiles:
We run unit tests across a variety of our projects, however we only build using ndk-build for the moment - we have a few bugs open against the experimental Gradle plug-in that need to be resolved before we can supply Gradle build files.

The most simple project to take a look at in terms of unit tests is fplutil https://github.com/google/fplutil 

e.g https://github.com/google/fplutil/tree/master/libfplutil/unit_tests/test_stdio uses libfplutil_main and libfplutil_print alongside googletest to run a set of tests to validate our standard output redirector library (libfplutil_print) is working correctly.  libfplutil_main uses native_app_glue to provide a standard C/C++ entry point to the application which allows us to package each unit test into an APK.

I suggest taking working through http://google.github.io/fplutil/libfplutil_overview.html to get a sense of how to make this work for your project.

moxi:
Hi my apologies, I was a little bit disconnected, I'll definitely take a look on all the input information and update this PR with the latest versions, To be honest I still want to do the merge of the setup and the example I had of getting to work the unit tests in Java with gradle experimental and having JNI involved and it will probably good to have another PR for the C/C++ unit tests to complement this module, I'll let you guys know when it's ready. What are your thoughts @proppy ?

proppy:
@moxi let's make a list of what we want to see in the sample first, and then divide the work :)

I think we want to show:
a. how to unittest JNI functions (what the current PR does)
b. how to build, use and run a c++ testing framework (like googletest, see also #142)
c. how to mix JNI and c++ test with a unified test runner and results
d. how to run c++ test and display result from android studio

I'd love to see one sample cover all those aspect, @JoseAlcerreca @moxi @stewartmiles do you agree that's a good way forward?

@moxi: how about I send a PR for b. against your fork and we continue to iterate on this PR together?

stewartmiles:
sgtm

moxi:
Yes, I agree that sound like a good plan for me.

ip:
@stewartmiles Thanks, these are helpful examples

ggfan:
CMake support released today, branch changed ( master is cmake code, gradle-experimental is previous master ). feel free to re-submit to new master branch after re-based, thanks

PR #94, opened 2015-09-04
lint native audio

proppy:
Fix the following lint warning
```
Correctness
2	Warning Assert: Assertions
2	Warning InlinedApi: Using inlined constants on older versions
Internationalization
21	Warning HardcodedText: Hardcoded text
```


ggfan:
LGTM

PR #96, opened 2015-09-08
add run time permission request for Android-M

ggfan:
This is to fix the crash when running the application on Android -M ( API 23 ).


native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
```diff
@@ -155,12 +170,19 @@ public void onClick(View view) {
 
         ((Button) findViewById(R.id.pause_uri)).setOnClickListener(new OnClickListener() {
             public void onClick(View view) {
+                if(!permissionGranted) {
+                    // Log.e(TAG, "Error: requested permission not granted, please restart app and grant access");
```
proppy: maybe you could leave the log.

native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
```diff
@@ -155,12 +170,19 @@ public void onClick(View view) {
...
+                    return;
+                }
                 setPlayingUriAudioPlayer(false);
              }
         });
 
         ((Button) findViewById(R.id.play_uri)).setOnClickListener(new OnClickListener() {
             public void onClick(View view) {
+                if(!permissionGranted) {
+                    return;
```
proppy: add logging?

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -80,7 +90,10 @@ public boolean onOptionsItemSelected(MenuItem item) {
     }
 
     public void startEcho(View view) {
-        status_view.setText("StartCapture Button Clicked\n");
```
proppy: is that intentional to get rid of this message?

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -38,6 +46,8 @@ protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_main);
 
+        //get run-time permission...
+        requestResourcePermission();
```
proppy: maybe you could request for the permission the first time someone click on the button?

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -16,17 +16,25 @@
...
 import android.view.MenuItem;
 import android.view.View;
 import android.widget.TextView;
 
-public class MainActivity extends Activity {
+public class MainActivity extends Activity
+        implements ActivityCompat.OnRequestPermissionsResultCallback {
     public static final String AUDIO_SAMPLE = "AUDIO_SAMPLE:";
+    private static final int AUDIO_ECHO_REQUEST = 0;
+    private Boolean permissionGranted = true;
```
proppy: should it be true by default?

native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
ggfan: not sure about this one: to follow the tradition for this sample, I should not add any log: original author removed all of the logs before he checked in code [ he left his TAG there commented out for later use I guess ]. So I followed his tradition to leave some hint there but not enabled it.

native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
ggfan: See the above explanation

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -120,11 +133,46 @@ private void queryNativeAudioParameters() {
...
     private void updateNativeAudioUI() {
-        status_view.setText("nativeSampleRate    = " + nativeSampleRate + "\n" +
-                "nativeSampleBufSize = " + nativeSampleBufSize + "\n" +
-                "nativeSampleFormat  = " + nativeSampleFormat);
+        if(permissionGranted) {
+            status_view.setText("nativeSampleRate    = " + nativeSampleRate + "\n" +
+                    "nativeSampleBufSize = " + nativeSampleBufSize + "\n" +
+                    "nativeSampleFormat  = " + nativeSampleFormat);
+        } else {
+            status_view.setText("Error: Resource Permission not Granted" +
```
proppy: it sounds that there is a `shouldShowRequestPermissionRationale()` function, that could help determine if the user has denied the permission previously (and if you need to show something informative about it).

See  https://developer.android.com/preview/features/runtime-permissions.html#explain-need

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: yeah, this is intentional: to prompt a button clicked may not very useful.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -120,11 +133,46 @@ private void queryNativeAudioParameters() {
...
+                                 "Please restart App and grant the necessary permissions");
+        }
+    }
+    private void requestResourcePermission() {
+        if (ActivityCompat.checkSelfPermission(this, Manifest.permission.RECORD_AUDIO) !=
+                PackageManager.PERMISSION_GRANTED) {
+            ActivityCompat.requestPermissions(this, new String[]{
+                            Manifest.permission.RECORD_AUDIO,
+                            Manifest.permission.MODIFY_AUDIO_SETTINGS,
+                            Manifest.permission.WRITE_EXTERNAL_STORAGE },
```
proppy: do you actually need `WRITE_EXTERNAL_STORAGE`?

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: this is good one: the authorization is a callback. I am not sure whether the callback is in the same thread or in different thread:  if it is in different thread, then we have to include a waiting if put this request inside the button click action handler.  felt this is easier and by requesting the permission during the app launch: for this app, if recording is not granted, we really could not do anything at all.

proppy: yes, that's true.

I was mainly wondering if it would be confusing since we can't display any 'messages' to the user before we show the permission.

native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
proppy: makes sense, filed #98 so we remember to cleanup logging across all the samples later

native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
proppy: Makes sense.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
proppy: Ack.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
proppy: Did you test the behavior of the app when the user deny the permission?

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: yes, the debug utility is to write out to /sdcard/data directory.  It is useful to analyze openSL ES  playback and capture callback regularity

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: yes, it is true by default to be compatible with previous APIs

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: when permission request is needed and denied, the status text box is going to show: "Error: permission not granted"  from then on click button will have no action at all. User needs to restart the app and grant the permission

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
proppy: But does the audio-echo app write to /sdcard/data?  (it seems that ENABLE_LOG is commented, so maybe the permission should also be guarded by a similar flag that you could query from C++).

What is AndroidLog advantage over logcat for logging text or time?

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
proppy: Maybe we should hide the buttons then, or just show a quit button.

native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
```diff
@@ -308,6 +336,35 @@ protected void onDestroy()
...
+    }
+
+    @Override
+    public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions,
+                                           @NonNull int[] grantResults) {
+
+        /*
+         * if any permission failed, the sample could not play
+         */
+        if (AUDIO_ECHO_REQUEST != 0) {
```
proppy: wouldn't this be always true? (because `private static final int AUDIO_ECHO_REQUEST = 0;`)

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -120,11 +133,46 @@ private void queryNativeAudioParameters() {
...
+    }
+
+    @Override
+    public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions,
+                                           @NonNull int[] grantResults) {
+
+        /*
+         * if any permission failed, the sample could not play
+         */
+        if (AUDIO_ECHO_REQUEST != 0) {
```
proppy: wouldn't this be always true? (because private static final int AUDIO_ECHO_REQUEST = 0;)

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -38,6 +46,8 @@ protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_main);
 
+        //get run-time permission...
+        requestResourcePermission();
         status_view = (TextView)findViewById(R.id.statusView);
         queryNativeAudioParameters();
```
proppy: Shouldn't you only do that if the permission is granted?

With the current code path I understand `Boolean permissionGranted` is only set to false after onRequestPermissionsResult, but since this is async, I don't think you have any guarantee that the code will block until the permission is granted.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -120,11 +133,46 @@ private void queryNativeAudioParameters() {
...
+                                "could not record Audio\n"    +
+                                 "Please restart App and grant the necessary permissions");
+        }
+    }
+    private void requestResourcePermission() {
+        if (ActivityCompat.checkSelfPermission(this, Manifest.permission.RECORD_AUDIO) !=
+                PackageManager.PERMISSION_GRANTED) {
+            ActivityCompat.requestPermissions(this, new String[]{
+                            Manifest.permission.RECORD_AUDIO,
+                            Manifest.permission.MODIFY_AUDIO_SETTINGS,
```
proppy: it seems that MODIFY_AUDIO_SETTINGS is in the list of [normal permissions](https://developer.android.com/preview/features/runtime-permissions.html#normal), so I don't think you need to request it explicitly.

native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
ggfan: done

proppy:
/cc @jfschmakeit could you take a look a this change, I made a few comments but I'm no expert on permissions.

@ggfan would like to focus the review on `audio-echo` first, and then update the other samples.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
jfschmakeit: Use "boolean" primitive type.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
jfschmakeit: Recommended UX here would be to request in context, ie. when clicking the button.
I agree that this will potentially complicate the flow of the app, but how about this:
- Request in onClickListener of record button:
    - If permission is missing, request it
    - If permission has been granted, continue as usual
- In onRequestPermissionsResult callback:
   - If granted, continue as usual 
   - If denied, display a dialog explaining that permissions are required, with a button to re-request permission or close the app

I'd be happy for you to move the permission request into onCreate, that way you don't need to handle different flows through the app (and handle the state where you are missing permissions. You could then remove the permissionGranted flag.)

We have done exactly this for some of our other platform samples too. See the relevant code here: https://github.com/googlesamples/android-Camera2Basic/commit/241c6fac8113ddb5dffb4f349b5bb9f1bea93501

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
jfschmakeit: +1, Ideally this should be done just before the permission request. If you are restructuring this change based on my suggestion (see above), a call to shouldShowRequestRationale might not be necessary though.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: yes, MODIFY_AUDIO_SETTINGS and INTERNET  are removed in last checkin

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: will look at this with new code changes after studying the JF's recommendation. The key is at the start button: as along as we get permission before we execute "start button" actions, app will not crash; if no permission when user click "start button", no call should go into JNI code.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: yes, it is indeed a bug. it is fixed in lated code checking.  thanks

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: yes, that is corrected, latest code removed the MODIFY_AUDIO_SETTINGS & INTERNET. thanks

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: Yes, will follow the samples JF recommended and compare the implementation.  The necessary idea for this sample is to protect "start" button: when "Start button" is clicked, if have permission --> call into JNI; if no permission, do nothing [ definitely not calling into JNI ] to avoid crash

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: yes, this is fixed in latest code checkin.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: for timing issues, you could not logcat to pring out: it takes too much time. The timing is critical for audio applications, the best way is to log into memory buffer and flush to disk when buffer is full or app closing so writing file is not a factor to audio timing.  I found I could find the timing irregularity when this simple logging so I used it.  Printing out to standout takes long [ I do not have numbers for comparison though]

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: I will study the sample to make things inside onCreate. Mainly to code is to demo the audio audio on Native code, simplify the logic on Java side would be nice

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: yes, it should be boolean, should be fixed in next future checkin.

audio-echo/app/src/main/AndroidManifest.xml
```diff
@@ -1,10 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.google.sample.echo" >
+
     <uses-permission android:name="android.permission.RECORD_AUDIO"></uses-permission>
     <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS"></uses-permission>
-    <uses-permission android:name="android.permission.INTERNET"></uses-permission>
-    <!-- debug-writing file need external storage writing -->
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
```
hak: Do we need the permission?

ggfan: have a debugging dumping the message into a file @ /sdcard/data directory. it is disabled by default though.  I will remove this one here and make a comment inside the code and get it addressed there.

thanks

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -120,11 +138,36 @@ private void queryNativeAudioParameters() {
...
+                                           @NonNull int[] grantResults) {
+        /*
+         * if any permission failed, the sample could not play
+         */
+        if (AUDIO_ECHO_REQUEST != requestCode) {
+            super.onRequestPermissionsResult(requestCode, permissions, grantResults);
+            return;
+        }
 
+        for (int res:grantResults) {
```
jfschmakeit: Make this check one line if you are only expecting one result (if(grantResults.length == 1 && grantResults[0] == ...), no need for the loop if you are only checking one permission.

audio-echo/app/src/main/jni/debug_utils.h
```diff
@@ -25,6 +25,7 @@
  *  Write given data to a file as binary file. File name is
  *     "/sdcard/data/audio_%d", file_index++
  *  requirement: must have /sdcard/data already created on android device
+ *               and request permission for WRITE_EXTERNAL_STORAGE
```
jfschmakeit: Remove.
(At least you removed the permission from the AndroidManifest?)

native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
```diff
@@ -33,9 +37,12 @@
 import android.widget.Spinner;
 import android.widget.Toast;
 
-public class NativeAudio extends Activity {
+public class NativeAudio extends Activity
+        implements ActivityCompat.OnRequestPermissionsResultCallback {
 
     //static final String TAG = "NativeAudio";
+    private static final int NATIVE_AUDIO_REQUEST = 0;
+    private Boolean permissionGranted = true;
```
jfschmakeit: 'boolean'

native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
```diff
@@ -308,6 +336,34 @@ protected void onDestroy()
...
+
+        /*
+         * if any permission failed, the sample could not play
+         */
+        if (NATIVE_AUDIO_REQUEST != requestCode) {
+            super.onRequestPermissionsResult(requestCode, permissions, grantResults);
+            return;
+        }
+
+        for (int res:grantResults) {
```
jfschmakeit: Simplify into one call

native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
```diff
@@ -308,6 +336,34 @@ protected void onDestroy()
...
+            if (res != PackageManager.PERMISSION_GRANTED) {
```
jfschmakeit: How are you handling the permission denied option here? It looks like none of the buttons will function without the permission and there is no user feedback that this is the reason.
What about bringing up a dialog after denial from which the user can re-request the permission or close the app?

native-codec/app/src/main/java/com/example/nativecodec/NativeCodec.java
```diff
@@ -315,5 +326,33 @@ void useAsSinkForNative() {
...
+        /*
+         * if any permission failed, the sample could not play
+         */
+        if (NATIVE_CODEC_REQUEST!= requestCode) {
+            super.onRequestPermissionsResult(requestCode, permissions, grantResults);
+            return;
+        }
+
+        for (int res:grantResults) {
+            if (res != PackageManager.PERMISSION_GRANTED) {
```
jfschmakeit: See my comment above, you need to handle the permission denied case as well.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: yes, changed in next checkin

audio-echo/app/src/main/jni/debug_utils.h
ggfan: done.

native-audio/app/src/main/java/com/example/nativeaudio/NativeAudio.java
ggfan: you are looking at the old code from here.  Latest code already addressed this and you have reviewed above. No more review necessary for any code from here and down below  -- they are older version of the same code. It is kind of github thing -- unsmart thing :-)

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -80,7 +89,16 @@ public boolean onOptionsItemSelected(MenuItem item) {
     }
 
     public void startEcho(View view) {
-        status_view.setText("StartCapture Button Clicked\n");
+        int permissionStatus = ActivityCompat.checkSelfPermission(this,
```
jfschmakeit: No need to store this in a variable if you are only checking it again in the next line.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -120,11 +138,36 @@ private void queryNativeAudioParameters() {
...
+         * if any permission failed, the sample could not play
+         */
+        if (AUDIO_ECHO_REQUEST != requestCode) {
+            super.onRequestPermissionsResult(requestCode, permissions, grantResults);
+            return;
+        }
+
+        if (grantResults.length != 1  ||
+            grantResults[0] != PackageManager.PERMISSION_GRANTED) {
+            status_view.setText("Error: failed to get user permission for RECORD_AUDIO");
```
jfschmakeit: Let's make the error more explicit: "Error: RECORD_AUDIO permission was denied." or similar.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -120,11 +138,36 @@ private void queryNativeAudioParameters() {
...
+            Toast.makeText(getApplicationContext(),
+                          "This sample needs RECORD_AUDIO permission",
```
jfschmakeit: Move to a string resource before merging.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
jfschmakeit: Please add a few short comments here, maybe something like:
// RECORD_AUDIO permission was denied. Display a short message and return.

jfschmakeit:
Thanks for your changes, @ggfan !
The new flow in audio-echo looks great and is sufficient for our existing samples. Ping me when the rest of the samples are ready for review too.

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: yeah: the only reason I did it is because of the 80 character/line code review requirement which ,I think, is one of the dumbest things in the world :-(.  I am glad you pointed it out, thanks!

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: done

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
ggfan: done

native-media/app/src/main/java/com/example/nativemedia/NativeMedia.java
```diff
@@ -206,10 +214,19 @@ public void onNothingSelected(AdapterView parent) {
         // initialize button click handlers
 
         // Java MediaPlayer start/pause
-
+        final Activity appActivity = this;
```
jfschmakeit: Rather than defining a final Activity here and potentially holding onto a reference, refer to the Activty as 'NativeMedia.this' in the click listener.

native-media/app/src/main/java/com/example/nativemedia/NativeMedia.java
```diff
@@ -404,4 +428,41 @@ void useAsSinkForNative() {
...
+                grantResults[0] != PackageManager.PERMISSION_GRANTED) {
+            /*
+             * When user denied the permission, throw a Toast to prompt that READ_EXTERNAL_STORAGE
+             * is necessary;
+             * This application go back to the original state: it behaves as if the button
+             * was not clicked. The assumption is that user will re-click the "start" button
+             * (to retry), or shutdown the app in normal way.
+             */
+            Toast.makeText(getApplicationContext(),
+                    getString(R.string.NeedPermission),
```
jfschmakeit: You can reference string resources directly in Toast.makeText, no need to call 'getString': http://developer.android.com/reference/android/widget/Toast.html#makeText(android.content.Context, int, int)

native-codec/app/src/main/java/com/example/nativecodec/NativeCodec.java
```diff
@@ -316,4 +333,38 @@ void useAsSinkForNative() {
...
+                    .show();
+            return;
+        }
+
+        /*
+         * When permissions are granted, prompt user with a toast and request user to re-click
+         * start button
+         */
+        Toast.makeText(getApplicationContext(),
+                getString(R.string.PermissionGranted),
```
jfschmakeit: You can reference string resources directly in Toast.makeText, no need to call 'getString': http://developer.android.com/reference/android/widget/Toast.html#makeText(android.content.Context, int, int)

native-codec/app/src/main/java/com/example/nativecodec/NativeCodec.java
```diff
@@ -316,4 +333,38 @@ void useAsSinkForNative() {
...
+                grantResults[0] != PackageManager.PERMISSION_GRANTED) {
+            /*
+             * When user denied the permission, throw a Toast to prompt that READ_EXTERNAL_STORAGE
+             * is necessary;
+             * This application go back to the original state: it behaves as if the button
+             * was not clicked. The assumption is that user will re-click the "start" button
+             * (to retry), or shutdown the app in normal way.
+             */
+            Toast.makeText(getApplicationContext(),
+                    getString(R.string.NeedPermission),
```
jfschmakeit: You can reference string resources directly in Toast.makeText, no need to call 'getString': http://developer.android.com/reference/android/widget/Toast.html#makeText(android.content.Context, int, int)

native-media/app/src/main/java/com/example/nativemedia/NativeMedia.java
```diff
@@ -404,4 +428,41 @@ void useAsSinkForNative() {
...
+                    Toast.LENGTH_SHORT)
+                    .show();
+            return;
+        }
+        /*
+         * When permissions are granted, prompt user with a toast and request user to re-click
+         * start button
+         */
+        Toast.makeText(getApplicationContext(),
+                getString(R.string.PermissionGranted),
```
jfschmakeit: ditto, use the string resource directly.

native-media/app/src/main/java/com/example/nativemedia/NativeMedia.java
```diff
@@ -206,10 +214,19 @@ public void onNothingSelected(AdapterView parent) {
         // initialize button click handlers
 
         // Java MediaPlayer start/pause
-
+        final Activity appActivity = this;
         ((Button) findViewById(R.id.start_java)).setOnClickListener(new View.OnClickListener() {
 
             public void onClick(View view) {
+                // Request the READ_EXTERNAL_STORAGE permission
+                if (ActivityCompat.checkSelfPermission(appActivity, Manifest.permission.READ_EXTERNAL_STORAGE)
```
jfschmakeit: You could move this check and request into a dedicated method (returning a boolean to indicate whether the permission is missing and has been requested.)
It's the same check as in the other clickListener.

jfschmakeit:
LGTM from me, just a handful of small nits, but the flow looks good. Feel free to submit once you have addressed or ACK'ed those last remaining comments. Thanks, @ggfan !

proppy:
should we submit this? I'm happy to take care of the nits in a follow PR if needed.

ggfan:
yes, please take care of it, proppy.  thanks

proppy:
@jfschmakeit I'll create another PR, pull @ggfan commits and address your comments.

PR #101, opened 2015-09-14
bump gradle android plugin to 0.4.0

proppy:
We should merge this once it's not alpha anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: android
 sudo: false
 android:
   components:
+    - platform-tools
     - build-tools-23.0.0
     - android-23
     - extra-google-m2repository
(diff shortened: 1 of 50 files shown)


ggfan:
LGTM

proppy:
@ggfan not LGTM, the build is failing :)

proppy:
Changelog from @rschiu
Fixed issue with using jni code in experimental library plugin.
Allow platform version to be set separately from compileSdkVersion in experimental plugin.
Allow ABI specific configurations in a variant that contains multiple ABI.
An preview version of an standalone plugin for compiling just native code is now available.  It can be use to build application with Gradle, but support in Android Studio is not yet implemented.
Added support for dependencies on NDK plugin and shared object/static library files.

proppy:
@rschiu I bumped the dependencies to alpha5, let's see if the CI builds pass now.

rschiu:
LGTM

PR #115, opened 2015-10-09
hello-thirdparty: add manual step to run fetch-gpg-sdk

### LINKED ISSUE #89: third-party example does not download package in default build
In the third-party example, build fails in compiling the JNI code. The gpg_sdk package is not downloaded and unpacked before compiling the JNI code in the default build configuration. Manually executing the download task first allows the build to proceed, but it would be nice to have a dependency to cause this to happen automatically.

Tested with Android Studio 1.3.2

Thanks

proppy:
Fixes #89

/cc @claywilkinson @rschiu @gcasey

I'm not very familiar with how to set non-library-dependencies between module tasks, let me know if there is a better way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,8 @@ script:
     (! grep -n $'\t' */app/src/main/AndroidManifest.xml) | cat -t; test ${PIPESTATUS[0]} -eq 0
   # check that there is no trailing spaces in AndroidManifest    
   - |-
-    (! grep -E '\s+$' */app/src/main/AndroidManifest.xml) | cat -e; test ${PIPESTATUS[0]} -eq 0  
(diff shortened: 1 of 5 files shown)


gcasey:
This was the same workaround I used when I encountered #89 so it seems like a good change to explicitly document the need for manual download.

Here are my other notes in case its helpful for figuring out how the automated download could be implemented.

What I noticed is the library dependency occurred with java compilation but the NDK compilation is a task that occurs before java.  This resulted in no guarantee the download would happen before NDK compilation.  If there was a way to make that dependency tied to the NDK compilation task rather than the java compilation task I think the problem would be solved, but I dont know how to do that either.

proppy:
@gcasey my understanding after chatting with @claywilkinson is that for this to work we need proper multi module support for gradle-experimental. /cc @rschiu

proppy:
we should update appveyor to download and extract the lib too.

proppy:
PTAL

Issue #57, opened 2015-08-10
Add sample showing custom jniLibs w/ Android.mk


mnbittorrent:
Is there a way to specify a jniLibs folder so that other .so files may be copied into the .apk?

proppy:
@mnbittorrent are you referring to a `.so` file build from another project (with Android.mk)?

proppy:
Related #26

mnbittorrent:
Yes, potentially. More likely in my case it would be a .so file manually compiled with a clang or gcc toolchain.

proppy:
@mnbittorrent if you have the option to compile a static library, the workaround pointed out in #26 might be interesting, see https://github.com/googlesamples/android-ndk/issues/26#issuecomment-133238221

proppy:
Repurposing this issue to add a sample that show how to set jniLibs option.

ph0b:
If the .so files are already compiled and you only want these to be integrated into the APK, you can change the jniLibs directory this way:

    android.sources{
        main.jniLibs {
            source {
                srcDir 'src/main/libs'
            }
        }
    }

proppy:
@ph0b do you have a good idea for a sample that show that?

Maybe an hybrid sample that use gradle for the app, and Android.mk to build a .so for libcurl?

Related #45

ph0b:
two-libs from the ndk would be a perfect candidate as it can't be converted to the gradle only build yet.

proppy:
@ph0b good idea!

proppy:
@phob but IIRC two-libs is using a static library:
https://github.com/googlesamples/android-ndk/blob/b4700b9499e1c6becc1a76b53a8691ecdeb25173/two-libs/jni/Android.mk#L30

So that would be a better candidate for #45 
- libone: a static library build with `Android.mk`
- libtwo: a dynamic library build with gradle depending on libone with custom `ldFlags`
- app: dynamic load libtwo with a standard `jniLibs`

Here I think we want something like:
- libone: a shared library build with Android.mk
- libtwo: a dynamic library build with gradle
- app: dynamically load libone and libtwo with a custom `jniLibs`

What do you think?

ph0b:
in the case you mention you would have no dependencies between the libs ? Then I don't find it really practical. I would better see the two-libs exactly as it is on the current ndk, keeping their current Android.mk/Application.mk. With the the app that dynamically loads them. 

[...]

PR #116, opened 2015-10-09
port two-libs sample to gradle

### LINKED ISSUE #113: add sample for 0.4.0 dependencies management
/cc @Whiteseeker

proppy:
This show how to use the new 0.3.0 dependencies support.

I'd also like to show in the same sample how to depends on a .so and add it in the .apk, maybe the name 'two-libs' is not really relevant anymore.

Any thoughts?

/cc @rschiu @ggfan @Whiteseeker @ph0b 

Fixes #113 #57


dragostis:
This is really similar to what I want to achieve apart from the fact that the library doesn't support Java files.

Are the normal libraries also getting the `productFlavor` [options](http://stackoverflow.com/questions/32957274/add-productflavor-to-experimental-android-gradle-plugin-library)?

two-libs/lib/build.gradle
```diff
@@ -0,0 +1,43 @@
+apply plugin: "com.android.model.native"
+
+model {
+    android {
+        compileSdkVersion = 23
+        buildToolsVersion = "23.0.0"
```
rschiu: Only the compileSdkVersion is needed.  Anything else in the android{} block should be removed.

two-libs/app/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+        buildToolsVersion = "23.0.0"
+
+        defaultConfig.with {
+            applicationId = "com.example.twolibs"
+            minSdkVersion.apiLevel = 4
+            targetSdkVersion.apiLevel = 23
+        }
+    }
+
+    compileOptions.with {
```
rschiu: Do we still need the compileOptions?

proppy: Yes, otherwise it would failed if java8 is installed.

proppy:
I think we should refactor this into a proper `native library sample` rather naively porting `two-libs`.

@claywilkinson suggest the following a gradle project with 3 modules.
- a `NativeLibrary` module
- 1 `AppDynamicLibrary` module that depends on `NativeLibrary` with dynamic linkage
- 1 `AppStaticLibrary` module that depends on `NativeLibrary` with static linkage

koyach:
Yeah sounds good

proppy:
`hello-libs` does this and more now.

PR #121, opened 2015-11-11
Fixed error where the plasma does not animate on some devices.

xdaimon:
This ndk example has never worked on my Nexus 7 2013, although it works on my Galaxy S3 for some reason. I finally decided to try and fix it.
If on the Galaxy S3 I increment the time_ms variable to be near the value which the Nexus 7 assigns time_ms, the plasma will not animate. Likewise if on the Nexus 7 I decrement the time_ms variable to be near the value which the Galaxy S3 assigns time_ms, the plasma will animate.
So instead of using just any time spit out by clock_gettime(), I decided to use the time since start of program.
This is my first contribution to any code online ever ;) Hope it helps!

### DIFF
--- a/native-plasma/app/src/main/jni/plasma.c
+++ b/native-plasma/app/src/main/jni/plasma.c
@@ -379,7 +379,6 @@ struct engine {
     int animating;
 };
 
-static struct timespec now;
 static int64_t start_ms;
 static void engine_draw_frame(struct engine* engine) {
     if (engine->app->window == NULL) {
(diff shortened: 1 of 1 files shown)


xdaimon:
I signed it!

native-plasma/app/src/main/jni/plasma.c
```diff
@@ -379,6 +379,8 @@ struct engine {
     int animating;
 };
 
+static struct timespec now;
```
proppy: maybe move that in `engine_draw_frame`, this doesn't need to be global.

native-plasma/app/src/main/jni/plasma.c
```diff
@@ -462,6 +463,9 @@ void android_main(struct android_app* state) {
         init = 1;
     }
 
+    clock_gettime(CLOCK_MONOTONIC, &now);
```
proppy: also use another local here.

proppy:
@xdaimon looks great thanks for testing and fixing this "very old" sample! I just made a few comments.

xdaimon:
Yeah, no problem.

proppy:
LGTM merging.

proppy:
Thanks for your contribution!

PR #129, opened 2015-11-30
port hello-thirdparty to use new 0.6.0 native dependencies system

ph0b:
Hi,

Here is a summary of the changes:
- bumped Google Play Services version to 8.3.0 and Play Games lib version to 2.0
- fixed crash at launch
- bumped gradle version to 2.9
- migrated native dependency declaration to the new gradle system supported in gradle-experimental:0.6.0-alpha2 and cleaned up the app build.gradle accordingly
- added support for all the architectures supported by Google Play Game
  Services

Note: To make it work in Android Studio 2.0, I had to deactive Instant Run for the project.

### DIFF
--- a/hello-thirdparty/app/build.gradle
+++ b/hello-thirdparty/app/build.gradle
@@ -33,7 +33,7 @@ model {
         moduleName = "native-activity"
         platformVersion = 9 //same as minSdkVersion.apiLevel for better compatibility
         stl    = "c++_static"
-        ldLibs.addAll(["atomic", "log", "android", "EGL", "GLESv1_CM", "z"])
+        ldLibs.addAll(["atomic", "log", "android", "EGL", "GLESv1_CM", "z"]) //gpg depends on atomic, log and z
         abiFilters.addAll(["armeabi", "armeabi-v7a", "x86", "arm64-v8a"]) //filtering ABIs on the ones Google Play Games library is compatible with.
     }
 
--- a/hello-thirdparty/app/build.gradle
+++ b/hello-thirdparty/app/build.gradle
@@ -8,7 +8,11 @@ model {
         libs(PrebuiltLibraries) {
             gpg {
                 headers.srcDir "${gpg_sdk_path}/include"
-                binaries.withType(SharedLibraryBinary) { // NOTE: this block should be "StaticLibraryBinary"/staticLibraryFile - but SharedLibraryBinary works and StaticLibraryBinary doesn't as of 0.6.0-alpha1
+                binaries.withType(SharedLibraryBinary) {
+                    //
+                    // NOTE: this block should be "StaticLibraryBinary"/staticLibraryFile - but SharedLibraryBinary works and StaticLibraryBinary doesn't as of 0.6.0-alpha2
+                    // bug reported here: https://code.google.com/p/android/issues/detail?id=196065
+                    //
                     sharedLibraryFile = file("${gpg_sdk_path}/lib/c++/${targetPlatform.getName()}/libgpg.a")
                 }
             }
--- a/hello-thirdparty/app/build.gradle
+++ b/hello-thirdparty/app/build.gradle
@@ -1,12 +1,23 @@
 apply plugin: 'com.android.model.application'
 
-
 def gpg_sdk_path = file(project(':thirdparty:gpg-sdk').projectDir).absolutePath + "/gpg-cpp-sdk/android"
 
 model {
+
+    repositories {
+        libs(PrebuiltLibraries) {
+            gpg {
+                headers.srcDir "${gpg_sdk_path}/include"
+                binaries.withType(SharedLibraryBinary) { // NOTE: this block should be "StaticLibraryBinary"/staticLibraryFile - but SharedLibraryBinary works and StaticLibraryBinary doesn't as of 0.6.0-alpha1
+                    sharedLibraryFile = file("${gpg_sdk_path}/lib/c++/${targetPlatform.getName()}/libgpg.a")
+                }
+            }
+        }
+    }
+
     android {
         compileSdkVersion = 23
-        buildToolsVersion = "23.0.0"
+        buildToolsVersion = "23.0.2"
 
         defaultConfig.with {
             //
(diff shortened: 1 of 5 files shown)


ph0b:
cc @proppy

hello-thirdparty/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.8-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.9-all.zip
```
proppy: we should probably do that across all the sample in a separate changes.

hello-thirdparty/app/build.gradle
```diff
@@ -1,9 +1,20 @@
...
-
 def gpg_sdk_path = file(project(':thirdparty:gpg-sdk').projectDir).absolutePath + "/gpg-cpp-sdk/android"
 
 model {
+
+    repositories {
+        libs(PrebuiltLibraries) {
+            gpg {
+                headers.srcDir "${gpg_sdk_path}/include"
+                binaries.withType(SharedLibraryBinary) { // NOTE: this block should be "StaticLibraryBinary"/staticLibraryFile - but SharedLibraryBinary works and StaticLibraryBinary doesn't as of 0.6.0-alpha1
```
proppy: @rschiu any idea why?

What was the behavior w/ 0.4? (maybe the feature wasn't available yet?).

hello-thirdparty/app/build.gradle
```diff
@@ -18,42 +29,34 @@ model {
...
+        platformVersion = 9 //same as minSdkVersion.apiLevel for better compatibility
         stl    = "c++_static"
-        ldLibs.addAll(["log", "android", "EGL", "GLESv1_CM", "z"])
-        //
-        //  NOTE: since libgpg is also linked to the stl:
-        //  libc++ explicitly follows it in the list of libraries
-        //  to prevent unresolved stl symbols.
-        //
-        ldLibs.addAll(["gpg", "c++_static"])
+        ldLibs.addAll(["atomic", "log", "android", "EGL", "GLESv1_CM", "z"])
```
proppy: @rschiu  since we don't list gpg ldLibs explicitly anymore (when using dependencies), does that mean that the need of re-appending `c++_static` goes away?

hello-thirdparty/app/build.gradle
```diff
@@ -18,42 +29,34 @@ model {
...
+        }
+    }
+
     android.buildTypes {
         release {
             minifyEnabled = false
-            proguardFiles.add(file('proguard-rules.txt'))
-        }
-    }
-    android.productFlavors {
```
proppy: are the productFlavors automatically inferred from the abiFilter?

proppy:
Thanks for the changes, it looks good, I just want some minor confirmation from @rschiu before merging.

hello-thirdparty/app/build.gradle
proppy: Also 0.6.0-alpha2 went out on monday.

hello-thirdparty/gradle/wrapper/gradle-wrapper.properties
ph0b: As you want, but gradle-experimental:0.6.0+ requires gradle 2.9, so you can do that in a separate change only if you do it before migrating any sample to gradle-experimental:0.6.0+.

proppy: oh I meant the change from -bin to -all, not the change from 2.8 to 2.9 :)

hello-thirdparty/app/build.gradle
ph0b: the feature wasn't available in 0.4. I've just tried with 0.6.0-alpha2 and get the same error:
:app:linkArmeabi-v7aDebugNative-activitySharedLibrary                 
org.gradle.nativeplatform.internal.prebuilt.PrebuiltLibraryResolveException: Shared library link file not set for prebuilt library 'gpg'.
        at org.gradle.nativeplatform.internal.prebuilt.AbstractPrebuiltLibraryBinary$ValidatingFileSet.getFiles(AbstractPrebuiltLibraryBinary.java:99)

I'll report it.

hello-thirdparty/app/build.gradle
ph0b: Fortunately it does. I only had to add atomic to the dependencies, as gpg links to it. the sample compiled and ran fine with these changes.

proppy: I wonder if we should make it more explicit that atomic is a dep of gpg (and not a dep of the sample) by having a separate ldLibs.addAll call.

hello-thirdparty/app/build.gradle
ph0b: Flavors? Where we're going, there is no flavors! ;)

ABI-specific flavors were used here only because they were initially required to support native dependencies (and during the Android Dev Summit, someone even said that ABI-specific flavors were "deprecated" now).
If you still prefer to generate one APK per architecture from this sample, there is a blocker: https://code.google.com/p/android/issues/detail?id=192943

proppy: Thanks, I'll log a bug to update all the samples with this pattern.

hello-thirdparty/app/build.gradle
ph0b: gpg depends on atomic, log and z (according to gpg lib's original sample Android.mk).
and the rest of the code depends on log, android, EGL and GLESv1_CM.
Since some dependencies are common (log), having a separate call to addAll with all the deps would lead to add these libs twice.
I've added a comment to explicit the dependency. We can have a separate call to add atomic and z too if you prefer.

proppy: Comment is perfect. Thanks!

hello-thirdparty/app/build.gradle
proppy: Can you add a link to the bug as a comment?

ph0b: done!

proppy:
I was tempted to wait for 0.6.0 to be non alpha before merging this, but since the same is already broken with 0.4.0, I think we can merge that now?

proppy:
Merging this now, since it crashed with the current sample.

proppy:
@ph0b it crashes for me on 5x, which device did you test with?

ph0b:
@proppy Zenfone 2.

I think the crash you're getting now is because your app_id is wrong.

cf. Readme.md:
Set APP_ID in app/res/values/ids.xml to the ID your project in the Game services section of the Google Play console. NOTE: If you fail to set the app ID or package name correctly, the sample will crash or behave unexpectedly.

On another hand a wrong app_id shouldn't lead to such a cryptic crash, I think we can add a new issue to the tracker for this :)

proppy:
Yes, make sense. @ggfan tested it successfully on M + Nexus 9.

Filed https://github.com/googlesamples/android-ndk/issues/131.

PR #133, opened 2015-12-10
Fix native-plasma sample to work on recent device (android 4.+)

Whiteseeker:
I know it's a lot of code changes but the current native-plasma hasn't been update since 2010.
It currently doesn't work on device that have a framebuffer with a format like RGBA8888 which have all the device around me (samsung galaxy s2, nexus 6, nexus 6P, motorola E 1st generation, ...)
I decide to separate the renderer from the plasma.cpp because we now need 2 different renderer depending on the framebuffer format.
I also moved away from define and instead used static const and functions.

PS: I also added the screenshot for the sample


Whiteseeker:
This is an attemps to fix #134

native-plasma/app/build.gradle
```diff
@@ -19,7 +19,7 @@ model {
 
     android.ndk {
         moduleName = "native-plasma"
-        CFlags.add("-I${file("src/main/jni/native_app_glue")}".toString())
+        cppFlags.add("-I${file("src/main/jni/native_app_glue")}".toString())
```
proppy: maybe we can switch to 0.6 and use dependencies instead?

native-plasma/app/src/main/jni/plasma.c
```diff
@@ -1,503 +0,0 @@
-/*
```
proppy: did you git mv the file?

proppy:
should we go one step further and merge the two plasma sample together?

native-plasma/app/build.gradle
Whiteseeker: I'd rather not until the issue with instant run is fixed :/ But if you insist I'll :)

proppy: you're right we can always do this across all the sample once it hit stables.

native-plasma/app/src/main/jni/plasma.c
Whiteseeker: I have to admit I didn't, bu there is very few lines in common between plasma.c and plasma.cpp since I move the rendering part out of the plasma.cpp.

Whiteseeker:
I totally want to merge both, could we first accept the updated native-plasma and then I'll make another pull request to merge both ?

proppy:
> I totally want to merge both, could we first accept the updated native-plasma and then I'll make another pull request to merge both ?

Sure make sense, maybe create a new issue for this? and ref it in this PR description.

native-plasma/app/src/main/jni/plasma.cpp
```diff
@@ -0,0 +1,262 @@
...
+  double firstTime;
+  double lastTime;
+  double frameTime;
+
+  int firstFrame;
+  int numFrames;
+  FrameStats frames[MAX_FRAME_STATS];
+} Stats;
+
+static void stats_init(Stats *s) {
```
proppy: maybe convert them to a Stats class?

native-plasma/app/src/main/jni/plasma.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+  Stats stats;
+
+  int initialised;
+  int animating;
+};
+
+static int64_t start_ms;
+
+static void engine_draw_frame(struct engine *engine) {
```
proppy: maybe have those as method of the Engine type.

native-plasma/app/src/main/jni/plasma_renderer_rgb565.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+static uint16_t palette[PALETTE_SIZE];
+
+static uint16_t makeColor(int red, int green, int blue) {
+  return (uint16_t) (((red << 8) & 0xf800) |
+                     ((green << 2) & 0x03e0) |
+                     ((blue >> 3) & 0x001f));
+}
+
+void PlasmaRendererRGB565::init_palette() {
```
proppy: Maybe you could share that function and take the palette size as an arg, or maybe it a template parameter?

Whiteseeker: It could totally go in the parent class PlasmaRenderer as it's shared by
both derived class, what do you think ?

native-plasma/app/src/main/jni/plasma_renderer_rgb565.cpp
```diff
@@ -0,0 +1,147 @@
...
+}
+
+uint16_t PlasmaRendererRGB565::palette_from_fixed(Fixed x) {
+  if (x < 0) x = -x;
+  if (x >= FIXED_ONE) x = FIXED_ONE - 1;
+  int idx = fixed_frac(x) >> (FIXED_BITS - PALETTE_BITS);
+  return palette[idx & (PALETTE_SIZE - 1)];
+}
+
+void PlasmaRendererRGB565::init() {
```
proppy: if init_palette is shared, maybe this could be shared as well.

native-plasma/app/src/main/jni/renderer_rbga8888.cpp
```diff
@@ -0,0 +1,101 @@
...
+    palette[nn] = makeColor(0, 255 - jj, 255);
+  }
+
+  for (mm = nn; nn < PALETTE_SIZE; nn++) {
+    int jj = (nn - mm) * 4 * 255 / PALETTE_SIZE;
+    palette[nn] = makeColor(jj, 0, 255);
+  }
+}
+
+uint32_t PlasmaRendererRGBA8888::palette_from_fixed(Fixed x) {
```
proppy: maybe this could also take parameter and be shared.

Whiteseeker: Well now that I think about it. The palette format is different for both
renderer, and the way you store the color is different. The only way to
achieve that would be to have a template in the parent class, and a
specialisation of template for the makeColor function (because the way to
store the color depend on the format of the data, uint16 or 32). It starts
to sound over engineered for a sample, knowing that most people doesn't
know well template and even less template specialisation. What do you think
?

proppy: what about a static function that takes a ref to array as an arg?
```
template<typename T, int PALETTE_SIZE>
void init_palette(const T (&palette)[PALETTE_SIZE]);
```

That way you can just do?
```
init_palette(palette_);
```

Does that sound over-engineered?

proppy: Oh scratch that, you're right the `makeColor` also change. It's fine to duplicate then :)

native-plasma/app/src/main/jni/plasma_renderer.h
```diff
@@ -0,0 +1,84 @@
...
+  static const int FIXED_BITS;
+  static const Fixed FIXED_ONE;
+  static const int ANGLE_BITS;
+  static const Angle ANGLE_2PI;
+  static const Angle ANGLE_PI;
+  static const Angle ANGLE_PI2;
+  static Fixed angle_sin_tab[];
+};
+
+class PlasmaRendererRGBA8888 : public PlasmaRenderer {
```
proppy: maybe also add the palette member here?

native-plasma/app/src/main/jni/plasma_renderer_rgb565.cpp
```diff
@@ -0,0 +1,147 @@
...
+ */
+#include "plasma_renderer.h"
+
+// Set to 1 to optimize memory stores when generating plasma.
+#define OPTIMIZE_WRITES 1
+
+/* Color palette used for rendering the plasma */
+static const int PALETTE_BITS = 8;
+static const int PALETTE_SIZE = 1 << PALETTE_BITS;
+
```
ggfan: this palette things are for 565 format, not related to 32 bit (A888) format, correct? wondering it could get pushed and tucked inside only 565 format implementation. 

probably original person declare everything in one place to allow for inline so it is faster?  if we pushed it inside the class, it might get impact on performance ( this is for every pixel ). 

I like the way you think to tuck them into a class

ggfan:
let's re-purpose this one? a couple of points
 1) 888X should be the same as the 8888 for code wise.  Probably add that support into the same code path. I think, right now, it crashes when other formats ( not 565, not 8888 ) are coming to us
 2) for other formats, assert them; so in the future we will know it is new format added
 3) since we are going to cpp, might create 3 classes
        render :: fill_plasma()=0   for interface
        then 2 render classes implement the interface ( one for 565, one 888X/8888 )
        
       that way the fill_plasma() function could be used at the rendering place, so client side code might be more organized

you code base might need to be rebased, significant changes in the latest repo:
   1) build script format changed ( it is inside build.gradle )
   2) native glue is replace with a static lib inside the same project
you will see them once you re-try merge

This samples bears the same with bitmap-plasma, we could think of retiring the bitmap-plasma once this is the shape we desire: for people that want the jni implementation, hello-jni + native-plasma would provide the same solution. what do you think?

Whiteseeker:
I'll just close that one. The main aim was to have it working with 8888 format to make it work on devices that support that format. But requesting to use 565 as you did in the other pr also solve the issue, so no need for this one anymore.

PR #137, opened 2015-12-16
Add sample for native soft keyboard

### LINKED ISSUE #136: Create sample showing how to use soft keyboard
Please ensure JB, L, and M compatibility or flavors of the sample that are compatible with each.

Per discussion with @proppy.

proppy:
Native SoftKeyboard
=============
Native SoftKeyboard is an Android sample that show how to toggle a software keyboard from a [Native Activity](http://developer.android.com/reference/android/app/NativeActivity.html).

Fixes #136

Tested on M/Nexus 6P.

Screenshots
-----------
screenshot


native-softkeyboard/app/src/main/jni/main.cpp
```diff
@@ -0,0 +1,258 @@
...
+#include <android/log.h>
+#include <android_native_app_glue.h>
+
+#define LOGI(...) ((void)__android_log_print(ANDROID_LOG_INFO, "native-activity", __VA_ARGS__))
+#define LOGW(...) ((void)__android_log_print(ANDROID_LOG_WARN, "native-activity", __VA_ARGS__))
+
+/**
+ * Shared state for our app.
+ */
+struct engine {
```
proppy: should use a class instead

native-softkeyboard/app/src/main/jni/main.cpp
```diff
@@ -0,0 +1,258 @@
...
+    struct android_app* app;
+
+    EGLDisplay display;
+    EGLSurface surface;
+    EGLContext context;
+    int32_t width;
+    int32_t height;
```
proppy: are width and height used?

native-softkeyboard/gradle/wrapper/gradle-wrapper.propertiese
```diff
@@ -0,0 +1,6 @@
+#Wed Apr 10 15:27:10 PDT 2013
```
claywilkinson: Is this a typo in the name? should it be gradle-wrapper.properties?

proppy:
It would be nice to display the typed character using a font.

We should investigate usage of `freetype`, https://github.com/nothings/stb/blob/master/stb_truetype.h or workaround using a bitmap font.

SHwareSystemsDev:
Wouldn't it be easier to instance a LinearLayout programatically, holding a centered label TextView... The KeyEvent handler then sets the Text property to do the echo for each char and you get freetype and bitmap font support in the process, whatever the user's default font is formatted as. When the keyboard hides you set the text to "". Just a thought...

proppy:
@SHwareSystemsDev I wanted to demonstrate usage of the soft keyboard to drive displaying text from C++, if we end up displaying the text  usingr Android UI framework we might as well drop the JNI call and do everything in Java :).

SHwareSystemsDev:
Well, the technique would be just for an echo in the default font, and as an example of how to instance a View in C++ without a layout xml I think it has collateral merit. It looked like you wanted the capability with the least cruft for the bang. 

That said, for a balance of cruft and bang, if echo box gets put in the lower right of the layout there's still the rest of the client area to do getGlyphForCode / putGlyphToGlCanvas in the event handler, and for laughs can implement for the glyphs used:
int loadBdfFile(char *path, struct bdfFont **font);
based on spec 5005 at:
http://www.adobe.com/devnet/font.html

A null or 0 length path could reference a default font for ASCII, as static data of the library. Add a Send To... intent handler for the bdf extension, another technique worth an example imo, and if people want to define their own fonts in a text editor they can launch the app from a file browser. The bitblt ops for scale, skew, and rotate would benefit from being native code, I imagine, if used randomly each time a char hits a layout side for a simple moving char animation. As a proof of concept format, while missing some desirable features, it being text instead of multiple binary structures lends itself to code that's almost self-documenting for the load part.

proppy:
> an example of how to instance a View in C++ without a layout xml

I think it would be useful to also have something that show how to inflate an xml layout from C++ without any Java code. We could also demonstrate how to rely on simple databinding instead of JNI heavy procedural view modification for C++ interop.

Will experiment a little bit a ping back that PR for review.

Thanks for all the input.

native-softkeyboard/app/src/main/jni/main.cpp
```diff
@@ -0,0 +1,258 @@
...
+    JNIEnv *jni;
+    app->activity->vm->AttachCurrentThread( &jni, NULL );
+
+    jclass cls = jni->GetObjectClass(app->activity->clazz);
+    jmethodID methodID = jni->GetMethodID(cls, "getSystemService", "(Ljava/lang/String;)Ljava/lang/Object;" );
+    jstring service_name = jni->NewStringUTF("input_method");
+    jobject input_service = jni->CallObjectMethod(app->activity->clazz, methodID, service_name);
+
+    jclass input_service_cls = jni->GetObjectClass(input_service);
+    methodID = jni->GetMethodID(input_service_cls, "toggleSoftInput", "(II)V");
```
hak: Maybe it can use ANativeActivity_showSoftInput() instead?
(not sure if the API has a pitfall and it's intensionally using the JNI call. If it's the case, it should state it.)

native-softkeyboard/app/src/main/jni/main.cpp
```diff
@@ -0,0 +1,258 @@
...
+            if (engine->app->window != NULL) {
+                engine_init_display(engine);
+                engine_draw_frame(engine);
+            }
+            break;
+        case APP_CMD_TERM_WINDOW:
+            // The window is being hidden or closed, clean it up.
+            engine_term_display(engine);
+            break;
+    }
```
SHwareSystemsDev: Appears these cases should have handlers also, even if do_nothings for now. Is where width and height, above, have significance as the validation bounds of a new windowRect or contentRect, I'd think.

+    /**
+     * Command from main thread: the current ANativeWindow has been resized.
+     * Please redraw with its new size.
+     */
+    APP_CMD_WINDOW_RESIZED,
+
+    /**
+     * Command from main thread: the system needs that the current ANativeWindow
+     * be redrawn.  You should redraw the window before handing this to
+     * android_app_exec_cmd() in order to avoid transient drawing glitches.
+     */
+    APP_CMD_WINDOW_REDRAW_NEEDED,
+
+    /**
+     * Command from main thread: the content area of the window has changed,
+     * such as from the soft input window being shown or hidden.  You can
+     * find the new content rect in android_app::contentRect.
+     */
+    APP_CMD_CONTENT_RECT_CHANGED,
+
+    /**
+     * Command from main thread: the app's activity window has gained
+     * input focus.
+     */
+    APP_CMD_GAINED_FOCUS,
+
+    /**
+     * Command from main thread: the app's activity window has lost
+     * input focus.
+     */
+    APP_CMD_LOST_FOCUS,
+

SHwareSystemsDev:
@proppy
You're welcome... I agree inflating a layout is also useful, but think this is complex enough it deserves a sample of its own as a separate library, e.g. native_layout_glue, due to the overhead of handling property inheritance from both View defaults and ViewGroup parents. That's based on looking at how it's done in Java now. Adding interop with alternate widget frameworks like KDE and tcltk far from trivial, as source or sink of any data bindings for layout params.

Imo, the XML the IDE adds to an apk that is of use is the locale specific strings and names of any bitmaps or icons the manifest references. Getting those out of the app jar and unzipped into C specific structures or temp sqlite table files useful to speed up any full redraws, this app or others.

native-softkeyboard/app/src/main/jni/main.cpp
proppy: Done. Thanks!

hak: OMG, saw some SO post for the API (http://stackoverflow.com/questions/5864790/how-to-show-the-soft-keyboard-on-native-activity) and verified ANativeActivity_showSoftInput()
doesn't work at all!! (Tested on N6P, M).
We need to revert the change and add some comments not to use the API :(

ggfan:
CMake support is released today, branch changed ( master is cmake code, gradle-experimental is previous master ). feel free to re-submit to new master branch after re-based, thanks

PR #138, opened 2015-12-17
fix static linkage declaration in hello-thirdparty

ph0b:
The bug https://code.google.com/p/android/issues/detail?id=196065 has been resolved: the linkage type has to be declared at the dependency level too. I've fixed the code accordingly.


ggfan:
LGTM

PR #143, opened 2015-12-30
bump build-tools

proppy:
/cc @stewartmiles


proppy:
Self merging, because the version "23.0.0" is deprecated and may causes issue when importing sample from android studio.

PR #144, opened 2015-12-31
build gradle plugin to 0.6.0

proppy:
Staging this for testing, we should only merge when the `-alpha` suffix is dropped.

Issue #60, opened 2015-08-18
Compilation error while running native-media


zagannath:
While running native-media sample on Android Studio AL-1.4.1 preview2 Im getting below errors.

Executing tasks: [:app:generateArm7DebugAndroidTestSources, :app:generateArm7DebugSources, :app:assembleArm7Debug]

Configuration on demand is an incubating feature.
:app:preBuild UP-TO-DATE
:app:preArm7DebugAndroidTestBuild UP-TO-DATE
:app:prepareArm7DebugAndroidTestDependencies
:app:compileArm7DebugAndroidTestAidl UP-TO-DATE
:app:processArm7DebugAndroidTestManifest UP-TO-DATE
:app:compileArm7DebugAndroidTestRenderscript UP-TO-DATE
:app:generateArm7DebugAndroidTestBuildConfig UP-TO-DATE
:app:generateArm7DebugAndroidTestAssets UP-TO-DATE
:app:mergeArm7DebugAndroidTestAssets UP-TO-DATE
:app:generateArm7DebugAndroidTestResValues UP-TO-DATE
:app:generateArm7DebugAndroidTestResources UP-TO-DATE
:app:mergeArm7DebugAndroidTestResources UP-TO-DATE
:app:processArm7DebugAndroidTestResources UP-TO-DATE
:app:generateArm7DebugAndroidTestSources UP-TO-DATE
:app:preArm7DebugBuild UP-TO-DATE
:app:checkArm7DebugManifest
:app:prepareArm7DebugDependencies
:app:compileArm7DebugAidl UP-TO-DATE
:app:compileArm7DebugRenderscript UP-TO-DATE
:app:generateArm7DebugBuildConfig UP-TO-DATE
:app:generateArm7DebugAssets UP-TO-DATE
:app:mergeArm7DebugAssets UP-TO-DATE
:app:generateArm7DebugResValues UP-TO-DATE
:app:generateArm7DebugResources UP-TO-DATE
:app:mergeArm7DebugResources UP-TO-DATE
:app:processArm7DebugManifest UP-TO-DATE
:app:processArm7DebugResources UP-TO-DATE
:app:generateArm7DebugSources UP-TO-DATE
:app:copyArmeabi-v7aDebugArm7Native-media-jniSharedLibraryGdbServer UP-TO-DATE
:app:createArmeabi-v7aDebugArm7Native-media-jniSharedLibraryGdbsetup
[...]

Issue #52, opened 2015-08-05
compiler error


CaiHuan:
UNEXPECTED TOP-LEVEL EXCEPTION:
com.android.dx.cf.iface.ParseException: bad class file magic (cafebabe) or version (0034.0000)
	at com.android.dx.cf.direct.DirectClassFile.parse0(DirectClassFile.java:472)
	at com.android.dx.cf.direct.DirectClassFile.parse(DirectClassFile.java:406)
	at com.android.dx.cf.direct.DirectClassFile.parseToInterfacesIfNecessary(DirectClassFile.java:388)
	at com.android.dx.cf.direct.DirectClassFile.getMagic(DirectClassFile.java:251)
	at com.android.dx.command.dexer.Main.processClass(Main.java:704)
	at com.android.dx.command.dexer.Main.processFileBytes(Main.java:673)
	at com.android.dx.command.dexer.Main.access$300(Main.java:83)
	at com.android.dx.command.dexer.Main$1.processFileBytes(Main.java:602)
	at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:170)
	at com.android.dx.cf.direct.ClassPathOpener.processDirectory(ClassPathOpener.java:229)
	at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:158)
	at com.android.dx.cf.direct.ClassPathOpener.processDirectory(ClassPathOpener.java:229)
	at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:158)
	at com.android.dx.cf.direct.ClassPathOpener.processDirectory(ClassPathOpener.java:229)
	at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:158)
	at com.android.dx.cf.direct.ClassPathOpener.processDirectory(ClassPathOpener.java:229)
	at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:158)
	at com.android.dx.cf.direct.ClassPathOpener.process(ClassPathOpener.java:144)
	at com.android.dx.command.dexer.Main.processOne(Main.java:632)
	at com.android.dx.command.dexer.Main.processAllFiles(Main.java:510)
	at com.android.dx.command.dexer.Main.runMonoDex(Main.java:280)
	at com.android.dx.command.dexer.Main.run(Main.java:246)
	at com.android.dx.command.dexer.Main.main(Main.java:215)
[...]

Issue #40, opened 2015-07-21
native-codec sample is buggy


proppy:
it crashes when pressing 'start' then 'rewind':
```
D/HeadsetStateMachine(22751): Disconnected process message: 10, size: 0
V/NativeCodec(32382): @@@ setsurface 0xb497ff08
V/NativeCodec(32382): @@@ create
V/NativeCodec(32382): opening /sdcard/testfile.mp4
V/NativeCodec(32382): failed: -1 (No such file or directory)
D/audio_hw_primary(  357): out_set_parameters: enter: usecase(1: low-latency-playback) kvpairs: routing=2
V/NativeCodec(32382): @@@ rewind
--------- beginning of crash
F/libc    (32382): Fatal signal 11 (SIGSEGV), code 1, fault addr 0xc in tid 32382 (ple.nativecodec)
D/audio_hw_primary(  357): out_set_parameters: enter: usecase(1: low-latency-playback) kvpairs: routing=2
I/DEBUG   (  354): *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
I/DEBUG   (  354): Build fingerprint: 'google/shamu/shamu:5.1.1/LYZ28H/1991770:user/release-keys'
I/DEBUG   (  354): Revision: '33696'
I/DEBUG   (  354): ABI: 'arm'
I/DEBUG   (  354): pid: 32382, tid: 32382, name: ple.nativecodec  >>> com.example.nativecodec <<<
I/DEBUG   (  354): signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0xc
I/DEBUG   (  354):     r0 0000000c  r1 a1b71df0  r2 00000000  r3 0000000c
I/DEBUG   (  354):     r4 0000000c  r5 12ce8d00  r6 12ce8e40  r7 bea1e1f8
I/DEBUG   (  354):     r8 12ce8e40  r9 b4827800  sl 00000000  fp 00000000
I/DEBUG   (  354):     ip b3683f90  sp bea1e1e8  lr b367ca01  pc b6e7e7ba  cpsr 000f0030
I/DEBUG   (  354): 
I/DEBUG   (  354): backtrace:
I/DEBUG   (  354):     #00 pc 000527ba  /system/lib/libc.so (sem_wait+19)
[...]

PR #148, opened 2016-02-01
Remove incorrectly set compileOpions

rschiu:


proppy:
IIRC that was done to make sure java7 is selected on a host os that also have java8.

Should we move this instead under the android block?

proppy:
`app:dexArm7Debug` was failing with java8, is that fixed now?

rschiu:
It should default to java 6.  So it would still work for java 8.  I'll double check to make sure.  We shouldn't set compileOptions for the users.

proppy:
>  We shouldn't set compileOptions for the users.

Yes, this was more of a temporary workaround, because multiple reports came up with this issue and we wanted the sample to work out of the box.

See #52 #34 #40 #60.

rschiu:
I tested it works with hello-jni arm7Debug.  Which project do you mean for app:dexArm7Debug?

proppy:
@rschiu that was the step failing when java8 is installed, see #52 #34

rschiu:
Ok, that is fixed now.  Please accept the pull request.  Thanks

proppy:
@rschiu, sure can you comment here on the commit/bug that fixed it? thanks.

proppy:
Note that #144 was also fixed the bad scoping of the `compileOptions`. I'll drop this change from it.

PR #150, opened 2016-02-02
Adding Activity fragment to support runtime permission check.

claywilkinson:
This creates an Activity Fragment that handles the flow of checking the runtime permissions.  The native code calls this with a callback which returns the result of the permission check.


claywilkinson:
@proppy, @ggfan - can you take a look at this PR?

proppy:
Does this replace #96?

claywilkinson:
Yes, I think so.  This is a more general solution that could easily be used in other native samples.

ggfan:
@proppy, may you have a look?  I might be look at this at the end of the week

audio-echo/app/build.gradle
```diff
@@ -30,3 +30,12 @@ model {
         create('all')
     }
 }
+configurations {
+    testCompile
```
proppy: is there some tests?

audio-echo/app/src/main/AndroidManifest.xml
```diff
@@ -1,18 +1,20 @@
...
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
+
     <application
         android:allowBackup="false"
         android:fullBackupContent="false"
         android:supportsRtl="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
-        android:theme="@style/AppTheme" >
+        android:theme="@style/Theme.AppCompat.Light" >
```
proppy: should we do that across all sample, if yes I'll file a bug.

proppy:
A few high level comments before diving more into this tomorrow.

- it would be nice to show how to do the fragment transaction in a c++ sample using native activity that doesn't have any Java code
- we should check that this is covering all the comments @jfschmakeit did in #96, maybe he can be ultra-generous and take the time to review this as well?

audio-echo/app/src/main/AndroidManifest.xml
claywilkinson: I think it is a good idea, but not critical.  There are a lot of new UI elements that require the appcompat library on older versions of Android.  I specifically changed this one for the UI elements used to display the permissions prompt and rationale.

claywilkinson:
Updated the build.gradle file.  OK to merge @proppy?

proppy:
@claywilkinson it would be cool if @jfschmakeit could proof read it before we merge to make sure it covers everything he reviewed as part of #96.

hak:
Clayton, could you resolve the conflict and re-submit the patch?

claywilkinson:
Abandoning this one and opened #189

PR #157, opened 2016-03-23
Update Teapot/MoreTeapot sample.

hak:
- Move ndk_helper and misc shared code under Common/
- Adjust viewport aspect when a screen is rotated.
- Make clang and libc++ as a default.
- Fix FPS disappearing after screen rotation.
- Clean up code.
- Update gradle versions in samples.
- Apply clang-format to files.

Tested: on Nexus6P with NDKr11.


ggfan:
LGTM.

PR #159, opened 2016-03-31
Fix the sample import feature in AndroidStudio.

hak:
- The feature requires all required source files inclusive in a folder.
- Duplicating Common files under Teapot/MoreTeapot samples as a temporary fix.
- Will re-organize the samples included in single project in future.

Tested: Built on OSX and samples work on N6P.


ggfan:
LGTM

PR #162, opened 2016-04-14
Converting native activity to C++

ggfan:
Be more strict when selecting config and print out openGL information
Update build.gradle to latest version


ggfan:
using another pull with latest source repo code

PR #163, opened 2016-04-14
Update native-activity

ggfan:
update native-activity to C++ app & 
using latest gradle DSL

### DIFF
--- a/native-activity/app/build.gradle
+++ b/native-activity/app/build.gradle
@@ -5,38 +5,39 @@ model {
         compileSdkVersion = 23
         buildToolsVersion = "23.0.2"
 
-        defaultConfig.with {
+        defaultConfig {
             applicationId = "com.example.native_activity"
             minSdkVersion.apiLevel    = 9
             targetSdkVersion.apiLevel = 23
         }
-    }
-
-    android.ndk {
-        moduleName ="native-activity"
-        CFlags.add("-I${file("src/main/jni/native_app_glue")}".toString())
-        ldLibs.addAll(["log", "android", "EGL", "GLESv1_CM"])
-    }
-    android.buildTypes {
-        release {
-            minifyEnabled = false
-            proguardFiles.add(file('proguard-rules.txt'))
+        ndk {
+            stl = "gnustl_static"
+            moduleName ="native-activity"
+            cppFlags.add("-I${file("src/main/jni/native_app_glue")}".toString())
+            cppFlags.add("-std=c++11")
+            ldLibs.addAll(["log", "android", "EGL", "GLESv1_CM"])
         }
-    }
-    android.productFlavors {
-        create ("arm7") {
-            ndk.abiFilters.add("armeabi-v7a")
+        buildTypes {
+            release {
+                minifyEnabled = false
+                proguardFiles.add(file('proguard-rules.txt'))
+            }
         }
-        create ("arm8") {
-            ndk.abiFilters.add("arm64-v8a")
+        productFlavors {
+            create ("arm7") {
+                ndk.abiFilters.add("armeabi-v7a")
+            }
+            create ("arm8") {
+                ndk.abiFilters.add("arm64-v8a")
+            }
+            create ("x86-32") {
+                ndk.abiFilters.add("x86")
+            }
+            // for detailed abiFilter descriptions, refer to "Supported ABIs" @
+            // https://developer.android.com/ndk/guides/abis.html#sa
+            // build one including all productFlavors
+            // build one including all productFlavors
+            create("fat")
         }
-        create ("x86-32") {
-            ndk.abiFilters.add("x86")
-        }
-        // for detailed abiFilter descriptions, refer to "Supported ABIs" @
-        // https://developer.android.com/ndk/guides/abis.html#sa
-        // build one including all productFlavors
-        // build one including all productFlavors
-        create("fat")
     }
 }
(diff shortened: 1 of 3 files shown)


native-activity/app/src/main/jni/main.cpp
```diff
@@ -85,10 +87,39 @@ static int engine_init_display(struct engine* engine) {
...
+        config = supportedConfigs[0];
+    }
+
+    /* EGL_NATIVE_VISUAL_ID is an attribute of the EGLConfig that is
+     * guaranteed to be accepted by ANativeWindow_setBuffersGeometry().
+     * As soon as we picked a EGLConfig, we can safely reconfigure the
+     * ANativeWindow buffers to match, using EGL_NATIVE_VISUAL_ID. */
+    eglGetConfigAttrib(display, config, EGL_NATIVE_VISUAL_ID, &format);
+
+    ANativeWindow_setBuffersGeometry(engine->app->window, 0, 0, format);
```
hak: It turned out this doesn't work on N5 (Broken screen happens by this).
>ANativeWindow_setBuffersGeometry(window, 0, 0, some_format);

Just remove the line would work.

native-activity/app/build.gradle
```diff
@@ -5,38 +5,39 @@ model {
...
-        release {
-            minifyEnabled = false
-            proguardFiles.add(file('proguard-rules.txt'))
-        }
-    }
-    android.productFlavors {
-        create ("arm7") {
-            ndk.abiFilters.add("armeabi-v7a")
+        ndk {
+            stl = "gnustl_static"
```
hak: (Just wondering) Is there a way to inherit stl flavor (and other setting if necessary) from the parent setting?

native-activity/app/build.gradle
```diff
@@ -5,38 +5,39 @@ model {
...
-        create ("x86-32") {
-            ndk.abiFilters.add("x86")
+        productFlavors {
+            create ("arm7") {
+                ndk.abiFilters.add("armeabi-v7a")
+            }
+            create ("arm8") {
+                ndk.abiFilters.add("arm64-v8a")
+            }
+            create ("x86-32") {
```
hak: IA32 ?

native-activity/app/build.gradle
```diff
@@ -5,38 +5,39 @@ model {
...
+                proguardFiles.add(file('proguard-rules.txt'))
+            }
         }
-        create ("x86-32") {
-            ndk.abiFilters.add("x86")
+        productFlavors {
+            create ("arm7") {
+                ndk.abiFilters.add("armeabi-v7a")
+            }
+            create ("arm8") {
```
hak: nit: arm64

native-activity/app/build.gradle
```diff
@@ -5,38 +5,39 @@ model {
...
+        buildTypes {
+            release {
+                minifyEnabled = false
+                proguardFiles.add(file('proguard-rules.txt'))
+            }
         }
-        create ("x86-32") {
-            ndk.abiFilters.add("x86")
+        productFlavors {
+            create ("arm7") {
```
hak: nit: arm32 ?

native-activity/app/build.gradle
```diff
@@ -5,38 +5,39 @@ model {
...
+                ndk.abiFilters.add("armeabi-v7a")
+            }
+            create ("arm8") {
+                ndk.abiFilters.add("arm64-v8a")
+            }
+            create ("x86-32") {
+                ndk.abiFilters.add("x86")
+            }
```
hak: Do we need IA64?

native-activity/app/src/main/jni/main.cpp
ggfan: done

native-activity/app/build.gradle
ggfan: maybe more explanation about parent project?  the build for each module inside the project is done separately ( right now ), so the parameters settings are not shareable.
for one module, we just have one stl_"gnustl_static" type.  it is not possible to share stl type between main project module and native-activity project (we put this one to a project later )

native-activity/app/build.gradle
ggfan: this abiFilter handling and split apk name thing might be addressed across board for all ndk samples:  they are all use the SAME one as here.  I think we might use next update to address them all, is that ok?

native-activity/app/build.gradle
ggfan: yes same as previous explanation, maybe next update to handle them all for ALL samples?

native-activity/app/build.gradle
ph0b: Why not removing the ABI specific flavors, at least for now ? 
The current solution isn't usable as-is since the version code of the generated APK is the same for each flavor, meaning only one of them can be distributed. 
In a next version, using splits and dynamic version codes could be much more useful - I'm not sure it's necessary to have it in every sample though.

ggfan: certainly removing all is an option. but what lost is the capability that I build one flavor (ABI) in a apk and sends that one via usb to the target system.  when all deleted, android studio will tuck every ABI in debug/release build, not sure we want that

ph0b: understood. Maybe you could switch now to a solution that works for everyone: splits (are they back in -experimental?) or flavors with dynamic version codes (https://code.google.com/p/android/issues/detail?id=192943)? If not, then I still think it's better with nothing, as at least the sample would be directly usable, even if means lengthier compilation times by default.

PR #173, opened 2016-04-22
Create hello-libs sample

ggfan:
Create hello-libs sample project:
       1) one application [uses 2 3rd party libs]
       2) 2 shared libs
       3) lib distribution dir is at ${project}/distribution/...
    
    Tested:
       Mac OS X + Simulator

Please kindly review and let me know your thoughts, thanks!
[Repo CI related thing is not addressed with this code change]


hello-libs/.gitignore
```diff
@@ -0,0 +1,8 @@
+*.iml
```
hak: Don't need the file?

hello-libs/README.md
```diff
@@ -0,0 +1,65 @@
+Hello-libs
+=========
+Hello-Libs is an Android sample that demos native lib management in Android Studio.
+
+This sample uses the new [Gradle Experimental Android plugin](http://tools.android.com/tech-docs/new-build-system/gradle-experimental) with external lib support:  
+*    Build JUST an native shared library[for distribution]
```
hak: nit 'an' -> 'a'

hello-libs/README.md
```diff
@@ -0,0 +1,65 @@
...
+*    Use 3rd party shared library / Static library   
+
+The key points are:
+*    Inform Android Studio where lib/header is
+*    Add lib dependency to application
+Details are inside projects' build.gradle files
+
+Sample contents:
+*    Create 2 shared libs
+*    Copy builts libs into distribution directory
```
hak: built

hello-libs/app/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
hak: don't need the file.

hello-libs/.gitignore
ggfan: no, this file is generated by android studio

hello-libs/README.md
ggfan: true: correct in next upload for sure. thanks

hello-libs/build_lib.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+pushd /Users/gfan/proj/android-ndk/hello-libs
+
+gradlew  :gmath:clean :gmath:assemble :gperf:clean :gperf:assemble
+
+mkdir  -p distribution/gmath/lib
```
hak: Wondering if we should copy files with gradle rather than sh?

hello-libs/gmath/src/main/jni/gmath.c
```diff
@@ -0,0 +1,47 @@
...
+#define GMATH_EXPORT
+#endif
+
+#define LOGE(...) \
+  ((void)__android_log_print(ANDROID_LOG_ERROR, "gmath::", __VA_ARGS__))
+
+/*
+ * return 2 ^ n with multiplication implementation
+ */
+GMATH_EXPORT unsigned power(unsigned n) {
```
hak: I understand the API is just a demonstration purpose.
But power() is in standard lib. Do we have any better example?

hello-libs/app/.gitignore
ggfan: true: removed

hello-libs/README.md
ggfan: yes, fixed in next

hello-libs/build_lib.sh
ggfan: deleted: it is not needed. I forgot to remove it. Gradle handles all

hello-libs/gmath/src/main/jni/gmath.c
ggfan: I think pow() is the standard function name. This is just something that simple enough to have something -- like a place holder. The goal is to show gradle's library processing, the content is not that significant.

PR #174, opened 2016-04-22
Retry pull request

ggfan:
Create hello-libs sample for review
  1) create one app
  2) create 2 shared libs ( to convert one to static later possible)
  3) use a common dir as distribution to simulate 3rd party


hak:
LGTM

ggfan:
thanks for review. I will merged it in: the CI error is again about the quotation thing, not related to this code change, need to learn the trick and just check version, ignore single/double quotation. context is here:
hello-thirdparty/app/build.gradle:        buildToolsVersion = "23.0.2"

PR #176, opened 2016-04-22
Removed hello-thirdparty: hello-libs replacing it

ggfan:
removed hello-thirdparty 
 add c++11 into hello-libs
  add android import related file into hello-libs


proppy:
LGTM

PR #178, opened 2016-04-25
Update plugin to 0.7.0-rc1 for clang build on windows(*) platform

ggfan:


ggfan:
clang could take inline prefix, it is also known issue  http://clang.llvm.org/compatibility.html#inline.  Clang throws out lot of errors for inline functions, so all -Werror flags are removed with this code checkin

bbilodeau:
LGTM

PR #181, opened 2016-04-29
minor enhancement to native-codec:

ggfan:
1) failed safe if no clip is on target
  2) move clips to app/src/main/clips/, and add jave code to move them
     to /sdcard/* during activity initialization


ggfan:
withdrawing this with a new approach: playing directly from asset folder

PR #182, opened 2016-04-29
Update for native-codec & native-media:

ggfan:
1) play from assets rather than from /sdcard
 2) if clip not found, mlooper will not be there, do not call rewind()
 3) update NativeMedia.ts to native-media

Tested: build on Mac, runs on Android-N/N5X
Known Issue: native-media native player still is not working


hak:
LGTM

PR #183, opened 2016-04-30
remove gpg reference since there is gpg in this game

ggfan:
No functional change, just wording in the about box.


bbilodeau:
LGTM

PR #187, opened 2016-05-03
Update platformVersion to all samples

ggfan:
1) update samples for platformVersion with minSdkVersion when possible; a couple exceptions which uses 64 bit print formatting, so it need to be 21 and above
2) change choreographer-30fps to use c++_static so it could compile with ndk-r12


hak:
LGTM

PR #188, opened 2016-05-04
change flag to mutex lock for native-audio

ggfan:


native-audio/app/src/main/jni/native-audio-jni.c
```diff
@@ -668,7 +675,7 @@ jboolean Java_com_example_nativeaudio_NativeAudio_selectClip(JNIEnv* env, jclass
         SLresult result;
         result = (*bqPlayerBufferQueue)->Enqueue(bqPlayerBufferQueue, nextBuffer, nextSize);
         if (SL_RESULT_SUCCESS != result) {
-            bqPlayerRecorderBusy = 1;
+            pthread_mutex_unlock(&audioEngineLock);
             return JNI_FALSE;
         }
```
hak: If Enqueue fails, is it still locked?

ggfan: the above code is unlock, so it is free for re-entry

hak: Had offline discussion and now I get the idea behind the change.

Just one point,
'bqPlayerRecorderBusy' was more descriptive than 'audioEngineLock' what's it's doing.
Can we do either:
- change 'audioEngineLock' name something like 'mutex_recorder_busy'?
or
- add some description what the mutex is for?

hak:
LGTM

PR #189, opened 2016-05-04
Adding Activity fragment to support runtime permission check.

claywilkinson:
Resubmitting PR 105 after resolution of conflicts.

This creates an Activity Fragment that handles the flow of checking the runtime permissions. The native code calls this with a callback which returns the result of the permission check.


ggfan:
For this case, app is audio playback / recording and will access disk, we could ask for the access at the startup of the app[that would be purely on Java side]; it is a fair statement that if user does not grant the access, the app will not enable record/play back UI button.  That way we save the trip to native and back to java UI traffic. If people are interested in how to use run-time permission, we could route them to samples that shows run-time permission; this way we keep native sample's jni side for purely audio purpose.

claywilkinson:
I think this change is more in line with being a native sample.  It moves more of the overall logic into C++; keeping on the the necessary code in Java (the fragment and permission lifecycle).    Is there a specific piece of code you are concerned about?

audio-echo/app/src/main/java/com/google/sample/echo/MainActivity.java
```diff
@@ -85,32 +95,13 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-            deleteSLBufferQueueAudioPlayer();
-            status_view.setText("Failed to create Audio Recorder");
-            return;
-        }
-        startPlay();   //this must include startRecording()
-        isPlaying = true;
-        status_view.setText("Engine Echoing ....");
+        status_view.setText("StartCapture Button Clicked\n");
+       startEchoProcessing();
     }
```
ggfan: would it be possible to do the permission at the place calling startEcho(): the picture would be --
   1) user wants to echo, click a button  ( on Java side )
   2) app requests permission ( on Java side )
   3) user grants permission ( on Java side )
   4) app calls startEcho()  to trigger jni to begin (on Java side ); after triggering, the work is done on JNI side for audio

if user does not grant the access, do not bother to call into jni at all, so it stays inside Java side to get work done

vs. right now:
    1) startEchoProcess   ( Java ==> JNI )
    2) calling back to Java for permission ( JNI ==> Java )
    3) Ask for permission ( JAVA )
    4) granted permission ( JAVA )
    5) call back to JNI ( JAVA ==> JNI )
    6) start echoing; from here it is on JNI side

ggfan:
tried to load on Nexus6 phone, after I denied the right, the app crashes. maybe have more look at it?

audio-echo/app/build.gradle
```diff
@@ -5,27 +5,32 @@ model {
...
+            applicationId = 'com.google.sample.echo'
             minSdkVersion.apiLevel = 17
-            targetSdkVersion.apiLevel = 22
+            targetSdkVersion.apiLevel = 23
             versionCode = 1
             versionName = '1.0'
         }
         ndk {
             // setting this to 21 because there is a printf using 64 bit formatting
             platformVersion = 21
```
ggfan: this one goes to 23, I think

audio-echo/app/src/main/AndroidManifest.xml
```diff
@@ -1,9 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.google.sample.echo" >
-    <uses-permission android:name="android.permission.RECORD_AUDIO"></uses-permission>
-    <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS"></uses-permission>
-    <uses-permission android:name="android.permission.INTERNET"></uses-permission>
+
```
ggfan: maybe this is extra line or something, I did not see reason this diff tool highlighting this line

audio-echo/app/src/main/AndroidManifest.xml
```diff
@@ -1,9 +1,10 @@
...
+    <uses-permission android:name="android.permission.RECORD_AUDIO"/>
+    <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS"/>
+    <uses-permission android:name="android.permission.INTERNET"/>
     <!-- debug-writing file need external storage writing -->
```
ggfan: INTERNET and STORAGE go away

audio-echo/app/src/main/java/com/google/sample/echo/PermissionRequestFragment.java
```diff
@@ -0,0 +1,167 @@
...
+                Log.w(TAG,permissions[0] + " denied");
+                callbackPtr = 0L;
+                handlePermissionResult(PackageManager.PERMISSION_DENIED, ptr);
+            }
+        } else {
+            super.onRequestPermissionsResult(requestCode, permissions, grantResults);
+        }
+    }
+
+    /*
```
ggfan: jni library load once is enough: it has been loaded from MainActivity, no need to re-load

audio-echo/app/src/main/java/com/google/sample/echo/PermissionRequestFragment.java
```diff
@@ -0,0 +1,167 @@
...
+                    view = thisActivity.getWindow().getDecorView();
+                }
+                Snackbar.make(view,
+                        rationale,
+                        Snackbar.LENGTH_INDEFINITE).setAction("OK", new View.OnClickListener() {
+                    @Override
+                    public void onClick(View view1) {
+                        theInstance.requestPermissions(new String[]{permission}, RC_REQUEST_PERMISSION);
+                    }
+                }).show();
```
ggfan: I found if I denied it, it crashes and I restarted it, it never asked me to re-try again. I guess this is the code causing the behaviour:
once user denied, user will not be asked again anymore.  She/He has to remove the application cache or remove app totally from the phone/tablet, re-install a new app, only then user could get a chance to re-asked for the permission.

I assume the samples are downloaded, built and run via android studio; user will touch "deny" to see the behaviour,  then user will find out that re-install fact, may be a little un-necessary step for development process

For formal release application, user will have to remove app cache or re-download from play store. 

is this intentional?

audio-echo/app/src/main/java/com/google/sample/echo/PermissionRequestFragment.java
```diff
@@ -0,0 +1,167 @@
...
+    @Override
+    public void onRequestPermissionsResult(int requestCode,
+                                           @NonNull String[] permissions, @NonNull int[] grantResults) {
+
+        if (requestCode == RC_REQUEST_PERMISSION) {
+            long ptr = callbackPtr;
+            if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
+                // finish starting
+                Log.d(TAG,permissions[0] + " granted!");
+                callbackPtr = 0L;
```
ggfan: nit:    
this is for locking purpose, I think.  maybe move it out of the if(){} else {} since it is common. Put it right after you made a copy of it on the top for easy code reading

ggfan:
CMake support is released today, branch changed ( master is cmake code, gradle-experimental is previous master ). Feel free to re-submit to the new master branch after re-based, thanks.

PR #190, opened 2016-05-05
Requesting 565 window format in native-plasma to make it work

ggfan:
this is to fix problem in: 
   https://github.com/googlesamples/android-ndk/issues/134

### DIFF
--- a/native-plasma/app/src/main/jni/plasma.c
+++ b/native-plasma/app/src/main/jni/plasma.c
@@ -427,15 +427,26 @@ static int32_t engine_handle_input(struct android_app* app, AInputEvent* event) 
 }
 
 static void engine_handle_cmd(struct android_app* app, int32_t cmd) {
+    static int32_t format = WINDOW_FORMAT_RGB_565;
     struct engine* engine = (struct engine*)app->userData;
     switch (cmd) {
         case APP_CMD_INIT_WINDOW:
             if (engine->app->window != NULL) {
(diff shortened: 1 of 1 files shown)


native-plasma/app/src/main/jni/plasma.c
```diff
@@ -427,15 +427,26 @@ static int32_t engine_handle_input(struct android_app* app, AInputEvent* event)
...
+                              WINDOW_FORMAT_RGB_565);
                 engine_draw_frame(engine);
             }
             break;
         case APP_CMD_TERM_WINDOW:
             engine_term_display(engine);
+            ANativeWindow_setBuffersGeometry(app->window,
+                          ANativeWindow_getWidth(app->window),
+                          ANativeWindow_getHeight(app->window),
+                          format);
```
hak: Do you need this line when terminating a window?
A window is going to be destroyed so my feeling is we don't need to restore it.

ggfan: yes, thought about that:  adding a few lines there to make sure we are a good citizen on the system and making no assumptions on what system is doing:
  system gives us a window, with some format
  system get back from us the same window with the same format
in case it is crashed somewhere and system wants to output some error message on that window in its final stage after calling _TERM_WINDOW, he could still do it. 

for this app, there is no effect whether we restore or not.

hak:
LGTM

PR #194, opened 2016-05-09
re-enable native player for native-media

ggfan:
Since M still works for native player, enable it and added a line in README.md to notify android-N issue. tracking with bug https://code.google.com/p/android/issues/detail?id=209410 and internal bug id 28666655


khirowatari:
LGTM. thanks.

PR #196, opened 2016-05-09
Fix issues/192

ggfan:
the default parameter swapped the order for up/low bound in Random()
  by mistake, this fixes it.

### DIFF
--- a/endless-tunnel/app/src/main/jni/util.hpp
+++ b/endless-tunnel/app/src/main/jni/util.hpp
@@ -27,7 +27,8 @@ template<typename T> void CleanUp(T** pptr) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


endless-tunnel/app/src/main/jni/util.hpp
```diff
@@ -27,7 +27,8 @@ template<typename T> void CleanUp(T** pptr) {
     }
 }
 
-int Random(int uboundExclusive, int lbound = 0);
+int Random(int uboundExclusive);
+int Random( int lbound, int uboundExclusive);
```
hak: nit: formatting.

hak:
except the nit spacing isdue, lgtm.

PR #197, opened 2016-05-10
adding debug related info

ggfan:
1) updated README.md to include hello-jni codelab which includes debug step inside Android Studio
2) added REFERENCE.md to put down gradle debugging flags

This is to solve https://github.com/googlesamples/android-ndk/issues/20
Plan: from here we add important gradle settings into REFERENCE.md, and hope someday it will be a thorough document. The format of the REFERENCE.md might change as we move along, the content should stay relevant.

Please review and comment whether this strategy could satisfy our needs, thank you very much!


bbilodeau:
LGTM

PR #198, opened 2016-05-10
light-weighted implementation for run-time permission request

ggfan:
affected audio-echo and native-audio samples


bbilodeau:
LGTM

PR #199, opened 2016-05-10
Revert platformVersion from Hello-Jni:

ggfan:
if version is too low (lower than the supporting for 64),
  all 64 bit ABI will not build and run anymore.

It is a big problem than just reverting this sample, need find a better way to handle it: one way or the other it has to break someone...


rschiu:
LGTM

PR #209, opened 2016-05-21
Create hello-jnicallback sample to complete hello-jni

ggfan:
With this sample, we complete the loop:
- calling from Java to C  (Hello-Jni)
- calling from C to Java  (this one)

It is simple sample: creating an independent thread and calling back to Java in every second. Functionality covered:
- create java obj in C
- call java static member function
- call java non-static member function
- call java private member function


hello-jniCallback/.gitignore
```diff
@@ -0,0 +1,8 @@
+*.iml
```
hak: Don't need the file.

hak: Don't need the .gitignore file.

hello-jniCallback/app/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
hak: Don't need the file.

hello-jniCallback/app/src/main/jni/hello-jnicallback.c
```diff
@@ -0,0 +1,314 @@
...
+    jstring buildVersion = (*env)->CallStaticObjectMethod(env,
+                                       g_ctx.jniHelperClz, versionFunc);
+    const char *version = (*env)->GetStringUTFChars(env, buildVersion, NULL);
+    if (!version) {
+        LOGE("Unable to get version string @ line %d", __LINE__);
+        return;
+    }
+    LOGI("Android Version - %s", version);
+    (*env)->ReleaseStringUTFChars(env, buildVersion, version);
+
```
hak: DeleteLocalRef to buildVersion ?

hello-jniCallback/app/src/main/jni/hello-jnicallback.c
```diff
@@ -0,0 +1,314 @@
...
+ * A helper function to wrap java JniHelper::updateStatus(String msg)
+ * JNI allow us to call this function via an instance even it is
+ * private function.
+ */
+void   sendJavaMsg(JNIEnv *env, jobject instance,
+                   jmethodID func,const char* msg) {
+    jstring javaMsg = (*env)->NewStringUTF(env, msg);
+    (*env)->CallVoidMethod(env, instance, func, javaMsg);
+    (*env)->DeleteLocalRef(env, javaMsg);
+}
```
hak: DeleteLocalRef to the javaMsg.

hello-jniCallback/app/src/main/jni/hello-jnicallback.c
hak: Added few comments, please take a look.

Do we want to have the sample in C rather than C++ style?
My feeling is for new samples, using C++ would be better for a readability and simplicity.

ggfan: DeleteLocalRef for buildVersion: maybe we do not need it here ( I am not totally sure )    
"Objects are passed to native methods as local references. All Java objects returned by JNI functions are local references. The JNI allows the programmer to create global references from local references. JNI functions that expect Java objects accept both global and local references. A native method may return a local or global reference to the VM as its result.

In most cases, the programmer should rely on the VM to free all local references after the native method returns "

Here "native method returns" means JNI_OnLoad() returns to system, I feel it is not needed to deleteLocalRef.

For C++ style, it is true it would be more concise with C++ which we do not need (*env)->(env,..) thing.  The major reason to keep this in C is that:  this sample meant to be  the second part of hello-jni, which is in C style. This one and hello-jni provide a round trip between java <--> native ( C ) code, and it might helpful to keep a migration trace from hello-jni to hello-jnicallback.  It could be either way, just for hello-jni purpose, it is kept in the similar style.

hak: As the thread is attached via AttachCurrentThread(), the VM won't release local references as described below. So we need to release it?

https://developer.android.com/training/articles/perf-jni.html
>One unusual case deserves separate mention. If you attach a native thread with AttachCurrentThread, the code you are running will never automatically free local references until the thread detaches. Any local references you create will have to be deleted manually. In general, any native code that creates local references in a loop probably needs to do some manual deletion.

I'm fine with the sample following C style now. In future we want to migrate all C samples to C++ ?

ggfan: that is good one, thanks!  maybe I do not need to use attachCurrentThread --     
 this function is called through JNI_OnLoad(), and everything has global reference is released in JNI_OnUnload().  The local reference is released there so system should clean it up there when system unloads shared lib?

I will remove attachCurrentThread from JNI_OnLoad(), seems it is still works w/o it, do you see any danger if I remove it?

ggfan: that is good point, JNI_OnLoad()/JNI_OnUnload(), those are called from JavaVM or system, it is not app's thread, correct? that is the reason the local reference is not released ( while everything inside app's new pthread are releases in that thread processing ) and we are piggybacking on system to do the work.
the question is whether I need to call attachcurrentthread() inside JNI_OnLoad(),  the thing seems still be working, so I am more liking to removing it from JNI_OnLoad() and let system takes care of that Load/Unload() pair local variables.

I think we need to keep a blend of C and C++ samples because JNI interface for these 2 are different, and android studio auto-generated prototype is C. Keeping hello-jni and this one  in C  to match up with android studio provides its unique value, I think.

ggfan: removed AttachCurrentThread() from JNI_OnLoad() code path; added DeleteLocalRef() for buildVersion. It is wired to be the minimum to have it in: which app does not create local ref, but need to release it?  it normally happens in pairs: one alloc, the same one release.  this is not the smart pointer in cpp thing...

hak:
LGTM.

Anu07:
Someone pls help!! letting me know where does cpu Info fetched from?

ggfan:
those are predefined gcc/clang macros: find where gcc ( or clang ) is run     
gcc -dM -E - <<<''  could get a list for that architecture, I think.

PR #215, opened 2016-05-26
remove JNI_OnUnload(): it never get called

ggfan:
also moved mutex manipulation with start/stop tick scope


khirowatari:
LGTM.

ggfan:
The :app:lint CI error is related to android studio 2.2 preview known issue,  merging in for now.

PR #216, opened 2016-05-26
removing comment about gradle.properties, no code changes

ggfan:
if we have sourceSets.main.jni.srcDirs = [], then gradle.properties is not needed


native-activity/app/build.gradle
```diff
@@ -26,7 +26,6 @@ android {
     // Documentation purpose: app/src/main/jni is reserved for deprecated native build system;
     // Steer away from this directory by DELETING it for ndk-build/cmake + android studio usage.
     // Otherwise you need to
```
rschiu: line format

rschiu:
LGTM

PR #217, opened 2016-05-27
remove hard float compiling flags: compiling flags are gone

ggfan:
this is to fix the issue: 
https://github.com/googlesamples/android-ndk/issues/207
there is no such compiling flags anymore in future ndk releases


hak:
LGTM

PR #230, opened 2016-06-13
fix hello-libs build issue

ggfan:
made two modules inside hallo-libs:
   1) gen-libs: to just produce the libs [within java lib module template]
   2) app:  use the generated libs

they are almost 100% independent: to save some writing, libs are FORCED to get re-built when app is building. They could really be decoupled, just keep them in one proj for easy maintenance. 

this would fix issue https://github.com/googlesamples/android-ndk/issues/226


hello-libs/app/src/main/cpp/hello-libs.cpp
```diff
@@ -21,6 +21,7 @@
 #include <gmath.h>
 #include <gperf.h>
 
+
```
hak: nit: extra space.

hak:
LGTM,

PR #232, opened 2016-06-14
splited hello-libs to 2 modules in the project, packed lib binaries

ggfan:


cmake/gen-libs/src/main/cpp/gmath/CMakeLists.txt
```diff
@@ -13,14 +13,19 @@
...
 cmake_minimum_required(VERSION 3.4.1)
 
 set(CMAKE_VERBOSE_MAKEFILE on)
 
 add_library(gmath STATIC src/gmath.c)
 
 # copy out the lib binary... need to leave the static lib around to pass gradle check
-set(distribution_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../distribution)
+set(distribution_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../../../../hello-libs/distribution)
 add_custom_command(TARGET gmath POST_BUILD
```
chaoren: You can use set_target_properties to output the library directly to the distribution location, and not have to copy and remove the original. See [here](https://github.com/ggfan/android-ndk/blob/1db95b23c74dd69ec61b6282f7c6bfd51217ef60/hello-libs/libsrc/gperf/CMakeLists.txt).

ggfan: done, good point! thanks

cmake/gen-libs/src/main/cpp/gmath/CMakeLists.txt
```diff
@@ -13,19 +13,25 @@
...
+                      ARCHIVE_OUTPUT_DIRECTORY
+                      "${distribution_DIR}/gmath/lib/${ANDROID_ABI}")
 add_custom_command(TARGET gmath POST_BUILD
                    COMMAND "${CMAKE_COMMAND}" -E
-                   copy "$<TARGET_FILE:gmath>"
-                   "${distribution_DIR}/gmath/lib/${ANDROID_ABI}/$<TARGET_FILE_NAME:gmath>"
-                   COMMAND "${CMAKE_COMMAND}" -E
                    copy "${CMAKE_CURRENT_SOURCE_DIR}/src/gmath.h"
                    "${distribution_DIR}/gmath/include/gmath.h"
 #                   **** the following 2 lines are for potential future debug purpose ****
```
chaoren: Looks like you can remove this comment now.

cmake/hello-libs/README.md
```diff
@@ -2,23 +2,23 @@ Hello-libs
 =========
 Hello-Libs is an Android sample that demos native lib management in Android Studio.
 
-This sample uses the new [Gradle Experimental Android plugin](http://tools.android.com/tech-docs/new-build-system/gradle-experimental) with external lib support:  
-*    Build JUST a native shared library[for distribution]
-*    Use 3rd party shared library / Static library   
+This sample uses the new [Gradle Experimental Android plugin](http://tools.android.com/tech-docs/new-build-system/gradle-experimental) with external lib support.
```
chaoren: This is actually Gradle stable plugin, not experimental.

cmake/gen-libs/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+# Generate one static lib and one shared lib, copy them into
+# ${project_dir}/distribution for other indepdendent applications
+# to use.
+cmake_minimum_required(VERSION 3.4.1)
+
+set(CMAKE_VERBOSE_MAKEFILE on)
+
+set(lib_src_DIR ${CMAKE_SOURCE_DIR})
+
+set(lib_build_DIR $ENV{HOME}/tmp)
```
chaoren: I don't know how people would feel about dumping a random directory in their home...

ggfan: yeah, this is mainly to show that we are really building in different directory and all temp files are removed after the build. So app surely could only use distribution location -- I like the temp directory to be far far away from app's normal build directory.  just to show that point, should be fine to create a directory at $HOME/tmp, most people having that already

cmake/hello-libs/README.md
ggfan: yes, I will update it

cmake/gen-libs/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+    buildToolsVersion "23.0.3"
+
+    defaultConfig {
+        minSdkVersion 13
+        targetSdkVersion 23
+        versionCode 1
+        versionName "1.0"
+
+        cmake {
+            abiFilters 'x86', 'x86_64', 'armeabi', 'armeabi-v7a', 'arm64-v8a', 'mips', 'mips64'
```
chaoren: You don't really need an ABI filter if you're gonna build everything.

chaoren:
LGTM.

PR #248, opened 2016-06-30
update to android studio 2.2.0 alpha5

ggfan:
It is not-backward compatible for cmake project:
    with android studio 2.2.0 alpha1 -- alpha4:  use code before this checkin
    with android studio 2.2.0 alpha5 and after:     this checkin and after


bbilodeau:
LGTM

PR #264, opened 2016-07-22
fixed issue https://github.com/googlesamples/android-ndk/issues/263:

ggfan:
using standard C++ headers, instead of c headers to fix PRIu64 issue


chaoren:
LGTM.

PR #274, opened 2016-08-05
Fix sample san-angeles for:

ggfan:
external build 'path' should be file
   add apk split feature
Branch:
   master-cmake
Tested:
   Mac X build, nexus 6 & emulator for target


tichang99:
LGTM

PR #279, opened 2016-08-16
Add a simple web psample

ggfan:


webp/.gitignore
```diff
@@ -0,0 +1,2 @@
+libwebp
```
hak: Do we need this file checked in?

webp/README.md
```diff
@@ -0,0 +1,53 @@
+WEBP Samples
+=============
+Webp is an Android sample that include a few sub apps to show usage of webp in  [Native Activity](http://developer.android.com/reference/android/app/NativeActivity.html).
```
hak: nit: that includes

webp/view/src/main/cpp/webp_decode.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include "webp_decode.h"
+
+WebpDecoder::WebpDecoder(const char** files, uint32_t count,
+                         DecodeSurfaceDescriptor* frameBuf,
+                         AAssetManager* assetMgr)
+    : assetMgr_(assetMgr),
+      buf_(nullptr), stopPending_(false),
+      state_(state_idle), worker_(nullptr) {
+    for (auto i = 0; i < count; i++)
+        files_.push(files[i]);
```
hak: We encourage putting braces all the time to avoid goto fail; bug.
https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/

for (auto i = 0; i < count; i++) {
}

webp/view/src/main/cpp/webp_decode.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    delete this;
+    return true;
+}
+
+/*
+ * private destructor prevent object directly call delete
+ */
+WebpDecoder::~WebpDecoder() {
+    if (buf_)  delete [] buf_;
```
hak: Probably we would prefer using std::unique_ptr to automatically delete them.

webp/view/src/main/cpp/webp_decode.h
```diff
@@ -0,0 +1,76 @@
+/*
```
hak: Indent issue. maybe 'clang-format -i --style=Google webp/view/src/main/cpp/webp_decode.h'

webp/view/src/main/cpp/webp_view.cpp
```diff
@@ -0,0 +1,280 @@
...
+    static int32_t format = WINDOW_FORMAT_RGB_565;
+    Engine* engine = reinterpret_cast<Engine*>(app->userData);
+    switch (cmd) {
+        case APP_CMD_INIT_WINDOW:
+            if (engine->AndroidApp()->window != NULL) {
+                format = ANativeWindow_getFormat(app->window);
+                ANativeWindow_setBuffersGeometry(app->window,
+                              ANativeWindow_getWidth(app->window),
+                              ANativeWindow_getHeight(app->window),
+                              WINDOW_FORMAT_RGB_565);
```
hak: Any reason we set to 565? If so, please add a comment.

webp/.gitignore
ggfan: yeah, mainly the libwebp directory is downloaded from github inside build.gradle, so we want to avoid it to be accidentally checked in.  If add to the upper layer directory's .gitignore,  I felt that later we will have to move it back whenever someone add a libwebp directory ( mainly I felt webp is going to be popular, more webp to come ).

webp/README.md
ggfan: fixed, thanks

webp/view/src/main/cpp/webp_decode.cpp
ggfan: ok, that is good to know. fixed

webp/view/src/main/cpp/webp_decode.cpp
ggfan: good point: in real application, those assert() would become "return" and will need a lot of delete.  The downside is that: gnustl_static does not support it, so need to switch to c++_static, which is the risky one for now.

webp/view/src/main/cpp/webp_decode.h
ggfan: I like to pardon this one, not like that tool to cut lines make code hard to read. sorry

webp/view/src/main/cpp/webp_view.cpp
ggfan: compack format, less copying. and it is also more problematic too, discovered that missed a flag to swap endianness for android.  32 is supported, works.

hak:
Thanks for fixes. LGTM.

PR #286, opened 2016-08-25
Resolve UnsatisfiedLinkError from bad abiFilter

ajgale:
The app crashes on launch with error:

java.lang.UnsatisfiedLinkError: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/com.example.hellolibs-1/base.apk"],nativeLibraryDirectories=[/data/app/com.example.hellolibs-1/lib/x86_64, /vendor/lib64, /system/lib64]]] couldn't find "libhello-libs.so"

Copying abiFilter creation from "Hello-JNI" app build.gradle resolves this issue.

### DIFF
--- a/hello-libs/app/build.gradle
+++ b/hello-libs/app/build.gradle
@@ -50,8 +50,33 @@ model {
             stl = 'gnustl_static'
             cppFlags.addAll(['-std=c++11'])
             ldLibs.addAll(['android', 'log'])
-            // build a default combined apk including all ABIs.
-            // abiFilters.addAll(['x86'])
+        }
+        productFlavors {
+            // for detailed abiFilter descriptions, refer to "Supported ABIs" @
+            // https://developer.android.com/ndk/guides/abis.html#sa
+            create("arm") {
+                ndk.abiFilters.add("armeabi")
+            }
+            create("arm7") {
+                ndk.abiFilters.add("armeabi-v7a")
+            }
+            create("arm8") {
+                ndk.abiFilters.add("arm64-v8a")
+            }
+            create("x86") {
+                ndk.abiFilters.add("x86")
+            }
+            create("x86-64") {
+                ndk.abiFilters.add("x86_64")
+            }
+            create("mips") {
+                ndk.abiFilters.add("mips")
+            }
+            create("mips-64") {
+                ndk.abiFilters.add("mips64")
+            }
+            // To include all cpu architectures, leaves abiFilters empty
+            create("all")
         }
         sources {
             main {


ajgale:
I signed it!

ggfan:
I thought it was building for an APK that including everything if not specified for anyone.  may you double check it?   it might be a different bug, I will also confirm it.  Thank you very much for finding the fix!

ajgale:
Hmmm... it looks like you're right that this is a different bug. I just ran it again with the original code and it works fine. So I created a fresh project from the sample and it ran fine again... So I created a new AVD and ran the new project on it... and it worked! I'll have a coworker try it with fresh instances of the project and an AVD tomorrow to see if he can reproduce the error I posted originally.

The first time I got it to work by uncommenting **build.gradle(Module:app) Line 54**:
`   // abiFilters.addAll(['x86'])
`

ggfan:
I am trying this: get a fresh code, change local.properties to point to my sdk and NDK r12, on terminal do a gradlew assembleDebug, it runs on my NEXUS 6 fine. I will do more to narrow down the cause: 
  1) directly build on command line
  2) let andrdoid studio open and update automatically and try again

kind of begin to guess latest android studio causes some conflict somewhere: this is also the only one we put a fake local.properties in the project... 
if you do not mind, let's do more poking at it and go from there. I think our patience should buy us something ( I am guessing again :-))

ggfan:
Ok, got some differences: may you try to see if it make any difference:
 download the repo, open Android Studio,  choose "Import project (Eclipse ADT, Gradle etc )" and
 - let android studio do the auto update  ( it should go to 0.8.0-beta2 )
 - then choose menu "Build" --> "Build APK" <=== this is important to me somehow
 - optional: "Build" --> "Analyze APK", double click to open "app-debug.apk, and expand "lib" folder, look through each sub ABI folder to see all 3 libs are there. I saw 7 ABIs are all there and all have 3 libs
 - Run  --> Run App to you device (or use the run button inside Android studio)

It seems more  consistent this way, may you check it out?

thanks

ajgale:
I'm seeing the same thing as you - all 7 ABIs are present in the APK and all have 3 libs. So far I have been unable to recreate the error.

The only other factor I can think of is that I updated to Studio Beta 2.2 yesterday and this was the first project I tried after updating, so maybe it has something to do with that? Happy to try any other suggestions you may have.

I'm updating another machine to the same configuration for testing and my coworker is also attempting to recreate so I will post our results.

ggfan:
I think it is more like a bug relating to Android studio, maybe instant run related. when I had error with Nexus 6,  it does build but build only one ABI: armeabi, which could not be used. It got to be something like: when I selected to "run" app, Android studio knows which device is to run, so Android Studio only builds for the target system ABI ( in this case it chooses armeabi which is wrong for my Nexus 6, armeabi-v7a would be good ).  When building for APK, android studio is not allowed to guess anymore so he has to follow the build.gradle's rule to build all ABIs ( that is our setting).
On command line/terminal, I tried pretty hard and never have this problem.

Again I think it is a bug: you found a bug in android studio.  Thank you!  what we will do is: get a consistent way to replicate and file a bug to tools.android.com ( most of my things are guessed info so far)

ajgale:
Good news! I've been able to consistently replicate the crash with the unmodified sample. 

Here are the steps:
1) Clean project
2) **Don't** build APK
3) Run app on any emulator that isn't x86
4) Crash with error "couldn't find 'libhello-libs.so' "

Analyze APK shows that only the lib/x86 folder contains "libhello-libs.so". 

I've submitted a bug report here: https://code.google.com/p/android/issues/detail?id=221050

PR #291, opened 2016-09-14
Update cmake samples to RC2 gradle plugin version

ggfan:
Also fix gen-libs for static lib build

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,5 @@ local.properties
 build
 *~
 .DS_Store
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -6,4 +6,5 @@ build
 *~
 .DS_Store
 externalNativeBuild
```
chaoren: You don't need this anymore.

ggfan: removed, thanks!

PR #292, opened 2016-09-15
Set event rate after enabling sensor so it takes effect.

jiawen:
The NDK documentation was not clear, but you are supposed to set the event rate
on an ASensorEventQueue only after enabling its associated sensor. It also needs
to be re-set after re-enabling the sensor. Without it, events arrive at some
"default rate" (quite slowly on some devices).


proppy:
LGTM

PR #298, opened 2016-09-21
Change to audio-echo to use shared lib 

ggfan:


audio-echo/app/build.gradle
```diff
@@ -15,8 +15,10 @@ android {
         }
         externalNativeBuild {
             cmake {
-                arguments '-DANDROID_STL=gnustl_static', '-DANDROID_TOOLCHAIN=gcc',
-                          '-DABDRIOID_PLATFORM=21'
+                // With android studio 2.2.0 +cmake, app needs to pack shared lib to
+                // into APK. Refer to CMakeLists.txt for details
+                arguments '-DANDROID_STL=c++_shared', '-DANDROID_TOOLCHAIN=gcc',
```
hak: Do we still want to use gcc? I think it's going to be deprecated.

yungangwang: '-DANDROID_STL=c++_shared' is right! thanks very much!

PR #299, opened 2016-09-22
Fix the build of the gles3jni sample

tberghammer:
This PR fixes http://b.android.com/223432


ggfan:
LGTM: thanks for the fix!

PR #303, opened 2016-10-05
Cleanup the cmake build files

tberghammer:
* Uniformalize the syntax used for the different projects
* Remove the usage of "glob" expressions
* Sort file lists to alphabetical order
* Replace uses of include_directories with target_include_directories
* Move some of the logic to subdirectories
* Remove a copy of native app glue


ggfan:
LGTM

PR #329, opened 2016-11-17
Fixed typo in in-line comments.

qchong:


ggfan:
LGTM, thanks

ggfan:
May you please get the same fix into more-teapots and choreographer-30fps too?  thank you so much!

PR #343, opened 2016-12-22
Consolidating ndk_helper source code

ggfan:
Tested: Nexus 9 ( android-23 ), nexus 5( android-25) and android-22
@hak PTAL, thanks


teapots/common/ndk_helper/JNIHelper.h
```diff
@@ -116,21 +160,61 @@ class JNIHelper {
...
    *
    * arguments:
    * in: file_name, file name to read, PNG&JPG is supported
+   * outWidth(Optional) pointer to retrieve original bitmap width
+   * outHeight(Optional) pointer to retrieve original bitmap height
+   * return:
+   * OpenGL texture name when the call succeeded
+   * When it failed to load the texture, it returns -1
+   */
+  uint32_t LoadTexture(const char* file_name, int32_t* outWidth = NULL,
```
hak: Prefer nullptr over NULL.

teapots/common/ndk_helper/JNIHelper.h
```diff
@@ -151,23 +235,31 @@ class JNIHelper {
   std::string GetExternalFilesDir();
```
hak: can be const?

teapots/common/ndk_helper/JNIHelper.h
```diff
@@ -151,23 +235,31 @@ class JNIHelper {
...
 
   /*
+   * Retrieve string resource with a given name
+   * arguments:
+   *  in: resourceName, name of string resource to retrieve
+   * return: string resource value, returns "" when there is no string resource with given name
+   */
+  std::string GetStringResource(const std::string& resourceName);
```
hak: can be const?

teapots/common/ndk_helper/JNIHelper.h
```diff
@@ -151,23 +235,31 @@ class JNIHelper {
...
    * Retrieves native audio buffer size which is required to achieve low latency
-   *audio
+   * audio
    *
    * return: Native audio buffer size which is a hint to achieve low latency
-   *audio
+   * audio
    * If the API is not supported (API level < 17), it returns 0
    */
   int32_t GetNativeAudioBufferSize();
```
hak: can be const?

teapots/common/ndk_helper/JNIHelper.h
```diff
@@ -178,6 +270,66 @@ class JNIHelper {
    *
    */
   const char* GetAppName() { return app_name_.c_str(); }
```
hak: can be const?

teapots/common/ndk_helper/JNIHelper.h
```diff
@@ -178,6 +270,66 @@ class JNIHelper {
...
+
+  /*
+   * Retrieves application label
+   *
+   * return: pointer to an app label string
+   *
+   */
+  const char* GetAppLabel() { return app_label_.c_str(); }
```
hak: can be const?

teapots/common/ndk_helper/perfMonitor.cpp
```diff
@@ -19,8 +19,10 @@
 namespace ndk_helper {
 
 PerfMonitor::PerfMonitor()
-    : tv_last_sec_(0), last_tick_(0.f), tickindex_(0), ticksum_(0) {
-  for (int32_t i = 0; i < NUM_SAMPLES; ++i) ticklist_[i] = 0;
+    : current_FPS_(0), tv_last_sec_(0), last_tick_(0.f), tickindex_(0),
+      ticksum_(0) {
+  for (int32_t i = 0; i < NUM_SAMPLES; ++i)
```
hak: Now we prefer kNumSamples style as a constant name.

teapots/common/ndk_helper/JNIHelper.h
ggfan: done

teapots/common/ndk_helper/JNIHelper.h
ggfan: maybe, but not sure it is necessary: your function is making std::string on the stack and returns this object ( it is not a pointer ). why guard against a temp object that is not shared with anyone? std::string s(path) is also making a copy of the internal chars in this string, it is NOT using the original memory from path and pointing to path's private char buffer.  I personally feel it is "nice" but may not be necessary: I would like to get waiver for it

teapots/common/ndk_helper/JNIHelper.h
ggfan: same as above: as long as you use code like:
  std::string s = std::string(resource);
  return s;

probably do not need to change the return type as const std::string.

teapots/common/ndk_helper/JNIHelper.h
ggfan: same as before.  and this is not purpose for this code change, so please waive it :-)

teapots/common/ndk_helper/JNIHelper.h
ggfan: same as before

teapots/common/ndk_helper/JNIHelper.h
ggfan: if you talking about make the function to be const as:
 std::string GetExternalFilesDir() const;
it will not compile as all functions you called inside this function has to have "const" character ( like AttachCurrentThread() etc )...

teapots/common/ndk_helper/JNIHelper.h
ggfan: changed it, nevertheless, to const char* GetAppLabel() const {return....}  does not feel there is any value added though: it is inline function and code is right there in the header: clients now it right away it does not change anything inside the class

teapots/common/ndk_helper/perfMonitor.cpp
ggfan: Ok, changed it. thanks

ggfan:
thanks for review, uploaded another patch addressing issues from code review.  please look at it again: like to get it in asap, thanks

teapots/common/ndk_helper/JNIHelper.h
hak: Agreed.

teapots/common/ndk_helper/JNIHelper.h
hak: Agreed.

hak:
LGTM.
Had a chat with gfan@ that when we want to add new sample using shared ndk_helper,
we might want to change either
- Change the name of teapots/
- Make ndk_helper as a submodule & change Android Studio import script to respect git submodules when making a package for a sample import in AS.

PR #344, opened 2016-12-24
fix(hello-libs): fix the EXT_LIB_ROOT point to wrong path which will raise compile failed

Jacksgong:


Jacksgong:
I signed it!

ggfan:
Oops, it was missed at the beginning of the port: thanks for the fix! LGTM

PR #353, opened 2017-02-05
Update SDK Version to 25

adithya321:


ggfan:
thank you for the Pull Request! Like to hold for this update to 2.3 - Beta3:  I think significant amount of people is still on the stable release, might be good to hold this update for a couple of more weeks.

ggfan:
the formal 2.3 is released: 
https://developer.android.com/studio/index.html

Earlier personally I experienced something strange with preview version, now we could update samples to formal 2.3 release.  May you kindly update it with the formal version and do some sanity check?  then we could merge the PR

thanks

adithya321:
Updated to 2.3.0.

ggfan:
LGTM

PR #363, opened 2017-02-22
Fix an outdated link in README.md

phillipao:
http://tools.android.com/tech-docs/external-c-builds has a redirector gadget. The gadget is broken, but the source has a reference to https://developer.android.com/studio/projects/add-native-code.html, so I guess that's the right place.


ggfan:
yes, thank you very much for the fix!  LGTM and merging it!

PR #368, opened 2017-03-10
fix OpenGL2 min android version

ggfan:
@qchong Please take a look, thanks


qchong:
LGTM, thanks for the updates!

PR #373, opened 2017-03-30
Use clang to build teapot samples

ggfan:
removing gcc: it is not supported
@bbilodeau @danalbert, please have review, thank you!


bbilodeau:
LGTM

PR #375, opened 2017-03-31
fix bug in webp sample for releasing memory and function return type

ggfan:
@bbilodeau @danalbert, please take a look at it, thank you so much!

### DIFF
--- a/webp/view/CMakeLists.txt
+++ b/webp/view/CMakeLists.txt
@@ -33,7 +33,7 @@ add_library(app-glue STATIC
 include_directories(${webp_src_dir}/examples
                     ${webp_src_dir}/src)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Werror")
 add_library(webp-view SHARED src/main/cpp/webp_decode.cpp
                              src/main/cpp/webp_view.cpp)
 
--- a/webp/view/src/main/cpp/webp_decode.cpp
+++ b/webp/view/src/main/cpp/webp_decode.cpp
@@ -63,6 +63,7 @@ uint8_t* WebpDecoder::GetDecodedFrame(void) {
  */
 void* DecodeFrame(void * decoder) {
     reinterpret_cast<WebpDecoder*>(decoder)->DecodeFrameInternal();
+    return nullptr;
 }
 
 /*
@@ -111,6 +112,7 @@ void WebpDecoder::DecodeFrameInternal() {
             break;
         default:
             assert( 0 );
+            delete  [] buf;
             return;
     }
     config.output.width = bufInfo_.width_;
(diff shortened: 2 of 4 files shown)


webp/view/CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ add_library(app-glue STATIC
 include_directories(${webp_src_dir}/examples
                     ${webp_src_dir}/src)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Werror -Wno-reorder")
```
DanAlbert: Why not just fix the `-Wreorder` issues?

webp/view/src/main/cpp/webp_decode.cpp
```diff
@@ -124,7 +125,7 @@ void WebpDecoder::DecodeFrameInternal() {
 
     status = WebPDecode(buf, len, &config);
     WebPFreeDecBuffer(&config.output);
-    delete buf;
+    delete [] buf;
```
DanAlbert: This is still leaked on L115. Just use `std::vector`.

webp/view/CMakeLists.txt
ggfan: that might be a too much a ask for developers to know the sequence of initialization: but it  definitely could be fixed.  Like to addressed together with other samples ( maybe this warning is in most samples I feel )

webp/view/src/main/cpp/webp_decode.cpp
ggfan: yeah, it is fixed with add back in delete [].  will look to make it real C++ like later thanks

ggfan:
please review it again, thanks!

webp/view/src/main/cpp/webp_decode.cpp
```diff
@@ -111,6 +112,7 @@ void WebpDecoder::DecodeFrameInternal() {
             break;
         default:
             assert( 0 );
+            delete  [] buf;
```
DanAlbert: Wait, if you're asserting this won't happen anyway then why not just `abort()`? This whole function is pretty suspicious. Most of these assertions failing mean something has gone horribly wrong and we cannot continue, but at best you just return after them and at worst continue as if nothing is wrong (`assert` will not abort in release builds unless you're explicitly passing `-UNDEBUG` when compiling, but you probably don't want to do that).

ggfan: sample is under debug build, assert should always happen; under release build the path is not addressed well: in some samples we are trying NOT to crash the things ( it might end up does nothing )

DanAlbert: Okay. This should probably all be cleaned up at some point, but it doesn't need to block this change. When it does get cleaned up, only use asserts for things where you do want to continue and explicitly abort when you can't.

webp/view/src/main/cpp/webp_decode.cpp
```diff
@@ -162,11 +164,15 @@ bool WebpDecoder::DecodeFrame(void) {
...
     pthread_attr_destroy(&attrib);
 
-    state_ = state_decoding;
-    return true;
+    if (status == 0) {
+        state_ = state_decoding;
+        return true;
+    }
+    // create thread failed...
+    assert(status == 0);
```
DanAlbert: This should still be addressed though.

webp/view/src/main/cpp/webp_decode.cpp
ggfan: agreed, will do it in next round of clean-up

PR #378, opened 2017-04-13
Update webp to clone libwebp repo as native code dependency

ggfan:
@bbilodeau @jomof @tjohns may you kindly look at it?  thank you


jomof:
What's the purpose of this change? Why not use .gitmodule?

ggfan:
the would decouple the cmake's dependency on gradle.

Android Studio "import module" does not work with .git submodule:  the submodule signature is inside repo's root directory.

jomof:
lgtm

PR #403, opened 2017-06-26
Add google maven repository declaration

Jose-Parente:
com.android.support:appcompat-v7 is no longer available
on the Android SDK off-line repository, it needs to be
downloaded from the google online repository instead


Jose-Parente:
As part of the release cycle of Android Studio, we are running some Automated tests that import the TeaPots sample on Android Studio, and then try to compile the project and run it on the emulator.
This is one of the problems we have at the moment.

PR #410, opened 2017-07-17
fixing issue in audio-echo

ggfan:
This fixes audio-echo issue reported from https://github.com/googlesamples/android-ndk/issues/367.
@dturner , @bbilodeau, PTAL


ggfan:
added native-audio's lock fix too.  @bbilodeau  PTAL

dturner:
Verified as fixed on Pixel running N2G47Q

PR #412, opened 2017-07-19
audio-echo: update UI, no functional change

ggfan:


ggfan:
Merging in for simple UI change

PR #416, opened 2017-07-25
Bump up Android level to 14 in application.mk

ggfan:
to make it buildable for NDK-r15 for Teapot and San-Angeles


ggfan:
Merging in for NDK-r15 test to pass

PR #418, opened 2017-07-28
Update build with Android Studio 2.3 + NDK-r15

ggfan:
Removed app_dummy() from NativeActivity derived samples
Note:
  NativeActivity derived samples will fail if NOT built with NDK-r15


ggfan:
Merging in since it is just build update: mainly to keep it for internal studio validation purpose.

PR #427, opened 2017-08-22
camera/basic sample: write file in a separate thread

ggfan:


ggfan:
merging for small change

PR #431, opened 2017-08-31
Camera: added GetLatestImage method

sjfricke:
As this being a very main source for anyone learning to use the Native Camera API calls it would be nice to show more information that if they plan to do real time processing that they should not be using `AImageReader_acquireNextImage` as it will cause huge delay when scaling up resolution of preview. Just adding the extra comment and function can help prevent searching and trying to debug why there is such a random image delay


ggfan:
LGTM, thx.

Next one is using OpenGL texture to convert it ( that will be done )

PR #442, opened 2017-10-26
destroy the thread attributes properly

rayworks:
The ```pthread_attr_t``` struct should be destroyed properly.


ggfan:
thank you for the PR!

rayworks:
;)

PR #443, opened 2017-10-26
Add missing includes to the teapots sample.

DanAlbert:
The newest NDK cleans up some accidental transitive includes, so decls
from these headers are no longer being leaked.


ggfan:
thanks for the PR: will go over others someday.  Vulkan samples having similar bug fix yesterday.
This sample has been there from the very beginning, and have been compiled wrong way?

PR #445, opened 2017-10-31
Update build for Android Studio 3.0

ggfan:
adding '[flavorDimensions](https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html?utm_source=android-studio#variant_aware)' for for hello-jni for Android Studio 3.0 requirements. @rschiu, @jomof, @qchong may you kindly take a look?

### DIFF
--- a/hello-jni/app/build.gradle
+++ b/hello-jni/app/build.gradle
@@ -2,7 +2,6 @@ apply plugin: 'com.android.application'
 
 android {
     compileSdkVersion 25
-    buildToolsVersion '25.0.2'
     defaultConfig {
         applicationId 'com.example.hellojni'
         minSdkVersion 23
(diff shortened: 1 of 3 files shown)


ggfan:
saw the reported issue: https://stackoverflow.com/questions/47023424/gadle-3-0-flavor-error-duplicate-class too.

hello-jni/app/build.gradle
```diff
@@ -2,7 +2,7 @@ apply plugin: 'com.android.application'
 
 android {
     compileSdkVersion 25
-    buildToolsVersion '25.0.2'
+    buildToolsVersion '26.0.2'
```
adarshf: Delete this line.

You no longer need to explicitly specify the build tools version. AGP uses a recommended default automatically :)

hello-jni/app/build.gradle
```diff
@@ -68,6 +76,6 @@ android {
 
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
-    compile 'com.android.support:appcompat-v7:25.2.0'
-    compile 'com.android.support.constraint:constraint-layout:1.0.1'
+    compile 'com.android.support:appcompat-v7:25.4.0'
```
adarshf: Could use global properties here?

See: https://developer.android.com/studio/build/gradle-tips.html#configure-project-wide-properties

(And yeah, I just noticed that I specify buildToolsVersion in the sample. I'll remove it :)

hello-jni/build.gradle
```diff
@@ -15,6 +16,7 @@ buildscript {
 allprojects {
     repositories {
         jcenter()
```
adarshf: Put google() first, before jcenter().

Believe it or not, certain things features, such as like running Lint checks on Kotlin code from the command line, doesn't work unless you include google() first (don't ask me why).

hello-jni/app/build.gradle
ggfan: this is good point, will use this global feature in multi-module project environment.  This one is one module project, keep it in one place for simplicity case.

hello-jni/app/build.gradle
ggfan: done

hello-jni/build.gradle
ggfan: done, will change other samples too

hello-jni/app/build.gradle
ggfan: done

PR #447, opened 2017-11-03
Teapot samples: bugfix for 439

ggfan:
This fixes https://github.com/googlesamples/android-ndk/issues/439

Tested: Samsung Galaxy S6 Edge / Android 7.0

### DIFF
--- a/teapots/choreographer-30fps/src/main/cpp/ChoreographerNativeActivity.cpp
+++ b/teapots/choreographer-30fps/src/main/cpp/ChoreographerNativeActivity.cpp
@@ -114,8 +114,8 @@ class Engine {
 
   Engine();
   ~Engine();
-  void SetState(android_app* state);
-  int InitDisplay();
+  void SetState(android_app *app);
+  int InitDisplay(android_app *app);
   void LoadResources();
   void UnloadResources();
   void DrawFrame();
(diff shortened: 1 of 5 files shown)


teapots/choreographer-30fps/src/main/cpp/ChoreographerNativeActivity.cpp
```diff
@@ -337,11 +337,19 @@ void Engine::UnloadResources() { renderer_.Unload(); }
...
   if (!initialized_resources_) {
     gl_context_->Init(app_->window);
     LoadResources();
     initialized_resources_ = true;
+  } else if(app->window != gl_context_->GetANativeWindow()) {
+    assert(gl_context_->GetANativeWindow());
+    UnloadResources();
+    gl_context_->Invalidate();
+    app_ = app;
+    gl_context_->Init(app->window);
```
hak: Below (switching line 348 & 349) would look little bit cleaner?

app_ = app;
gl_context_->Invalidate();
gl_context_->Init(app->window);

teapots/choreographer-30fps/src/main/cpp/ChoreographerNativeActivity.cpp
```diff
@@ -337,11 +337,19 @@ void Engine::UnloadResources() { renderer_.Unload(); }
...
  * Initialize an EGL context for the current display.
  */
-int Engine::InitDisplay() {
+int Engine::InitDisplay(android_app *app) {
   if (!initialized_resources_) {
     gl_context_->Init(app_->window);
     LoadResources();
     initialized_resources_ = true;
+  } else if(app->window != gl_context_->GetANativeWindow()) {
+    assert(gl_context_->GetANativeWindow());
```
hak: Can we have some comments here?

e.g.
Re-initialize ANativeWindow. On some devices, ANativeWindow is re-created when the app is resumed.

ggfan: done, thanks!

teapots/choreographer-30fps/src/main/cpp/ChoreographerNativeActivity.cpp
ggfan: Yeah, putting "app_ = app" after Invalidate() is safer:  when invalidating, whatever the resources allocated before are created with old window, let them released with the old window handle attached to them still; But we must put the "app" into cached "_app" right before any Init(), so in case there is any hidden usage of it, it is automatically handled with new window handle.  I felt it is safe this way: no need to worry too much about hidden usage of _app in other part of the code.

hak: LGTM

PR #452, opened 2017-11-18
Replaced deprecated display.getWidth() and display.getHeight()

TheScottyJr:


TheScottyJr:
I signed it!

ggfan:
thanks for the PR: please create more for other parts of code too. Java side of the code was not followed closely, your help is valuable!

Schumanzou:
I used the AndroidLog, but the fopen always return NULL.

Issue #450, opened 2017-11-13
hello-libs: Error with Gradle plugin 3.0.0 when re-building libs


joe-skb7:
I'm trying to make libs to be rebuilt on each build. For that I did next:
 1. Removed pre-built libraries (`distribution/` directory)
 2. Enabled `gen-libs` module building (commented out corresponding lines in `app/build.gradle` and `settings.gradle` files)

With current state of android-ndk project (commit 4df5a27), build works fine.

But after updating the project to use Gradle plugin 3.0.0 (in Android Studio 3.0), I'm seeing next error when running build:

    :app:externalNativeBuildDebug
    Build hello-libs arm64-v8a
    ninja: error: '.../hello-libs/distribution/gmath/lib/arm64-v8a/libgmath.a', needed by '.../hello-libs/app/build/intermediates/cmake/debug/obj/arm64-v8a/libhello-libs.so', missing and no known rule to make it
    :app:externalNativeBuildDebug FAILED
    
    FAILURE: Build failed with an exception.
    
    * What went wrong:
    Execution failed for task ':app:externalNativeBuildDebug'.
    > Build command failed.
      Error while executing process ~/Android/Sdk/cmake/3.6.4111459/bin/cmake with arguments {--build .../hello-libs/app/.externalNativeBuild/cmake/debug/arm64-v8a --target hello-libs}
      ninja: error: '.../hello-libs/distribution/gmath/lib/arm64-v8a/libgmath.a', needed by '../hello-libs/app/build/intermediates/cmake/debug/obj/arm64-v8a/libhello-libs.so', missing and no known rule to make it

Is there some dependency missing? As far as I understand, `libhello-libs.so` must depend on `libgmath.a` and `libgperf.so`, but we somehow don't have such a dependency. Can you fix it please?

**Supplementary 1**: when doing `Build -> Make Module 'gen-libs'` first, project build works fine. But this is just a workaround. We need to come up with proper dependency instead.

**Supplementary 2**: I found one possible solution to this problem (not sure if it's the best, though). Just added next code in the end of `app/build.gradle` and now build seems to work correctly:

```gradle
tasks.whenTaskAdded { task ->
    if (task.name == 'externalNativeBuildRelease') {
        task.dependsOn ":gen-libs:externalNativeBuildRelease"
    } else if (task.name == 'externalNativeBuildDebug') {
        task.dependsOn ":gen-libs:externalNativeBuildDebug"
    }
}
```

[...]

PR #455, opened 2017-11-22
hello-libs: Update to Gradle plugin 3.0.1 and fix collateral issues

joe-skb7:
Migrating hello-libs project to Android 3.0.1 and Gradle plugin 3.0.1 is breaking normal build. This patch series fixes all connected issues and updating Gradle plugin to v3.0.1.

Mentioned issues were reported in #450. This patch series fixes that.


hello-libs/app/build.gradle
```diff
@@ -42,10 +41,18 @@ android {
...
     // uncomment out this one to generate lib binaries,
     // and also uncommented out the one in settings.gradle
     // after lib is generated, just comment them out again
 //        compile project(':gen-libs')
 }
+
+tasks.whenTaskAdded { task ->
+    if (task.name == 'externalNativeBuildRelease') {
+        task.dependsOn ":gen-libs:externalNativeBuildRelease"
+    } else if (task.name == 'externalNativeBuildDebug') {
```
ggfan: thanks for the patch!  may you add:

- a line of comment to indicate that it is for building the lib only
- by default, I think we still want to build just the application, do not want to build the lib. maybe put the whole lib build into comment ( I will also update the readme.md to reflect it ).

To build app and lib ( and link at different location ) has some duplicate work: for shared libs, when building lib, it is packed to APK ( relating to jniLibs of AAR ); then when linking with app's  CMakeLists.txt ( import clause for shared lib), it will be packed again; the second time probably just overwrite the first version; not important, like to avoid pointing people to the wrong direction. kindly let me know what you think, thanks!

joe-skb7:
Sorry, I was really busy last days... I see you already merged it and added missing comments, thanks. The only thing I can add (the minor one) is that `hello-libs/app/build.gradle` file should have newline in the end. As I remember, Git requires it.

ggfan:
I will fix it (newline in the end issue), thank you very much!

PR #460, opened 2017-11-29
Fix wrong doc in hello-libs' README.md

ggfan:
Fix https://github.com/googlesamples/android-ndk/issues/459.  The documentation is wrong actually ( not just "failed to mention..." ). Fixed with PR,  if Studio's cmake port could directly pack it, that would be handy.
@mondaugen  please take a look.


ggfan:
@mwgray, may you kindly take a look? thanks

PR #461, opened 2017-11-30
add abspath workaround for windows ndk-build projects

ggfan:
tested: Windows & Mac OS

This is mainly to fix ndk-build under windows. The 240 is magic number for file path, abspath helps. refer to:
  http://gnu-make.2324884.n4.nabble.com/using-realpath-abspath-in-3-81-td13798.html
at the bottom of the thread is the workaround, works fine for this case.


synthesisGoogle:
LGTM

PR #462, opened 2017-12-01
Minor update gles3jni to use version 3.2 headers for API 24+

ggfan:
mainly to address issue: https://github.com/googlesamples/android-ndk/issues/350
There is validation usage for various ndk versions, but the sample content no change ( not exploring any of the new functionalities inside gles32)


ggfan:
@hak PLAT, thx

PR #466, opened 2017-12-08
Switching to c++_* stl and removing armeabi ABI

ggfan:


ggfan:
just to remove deprecated gnustl_* and armeabi, @DanAlbert, @jomof Please take a look

bitmap-plasma/app/build.gradle
```diff
@@ -8,7 +8,7 @@ android {
         minSdkVersion 14
         targetSdkVersion 25
         ndk {
-            abiFilters 'armeabi', 'armeabi-v7a', 'arm64-v8a', 'x86'
+            abiFilters 'armeabi-v7a', 'arm64-v8a', 'x86'
```
DanAlbert: Why not x86_64?

What happens if you don't set this? Won't it default to all? If not, we should change that.

camera/texture-view/build.gradle
```diff
@@ -12,7 +12,7 @@ android {
         }
         externalNativeBuild {
             cmake {
-                arguments '-DANDROID_STL=gnustl_static', '-DANDROID_TOOLCHAIN=clang'
+                arguments '-DANDROID_STL=c++_static', '-DANDROID_TOOLCHAIN=clang'
```
DanAlbert: Why are these set explicitly? Just let the cmake toolchain decide what's best for the given NDK. The only time you should need to set this explicitly is if you need the shared STL rather than the static STL, or if you're using a feature that is not available in gnustl (but that by definition is not the case for any of these, since they were already gnustl).

Same goes for `ANDROID_TOOLCHAIN`. Clang has been the default for ages.

bitmap-plasma/app/build.gradle
ggfan: It still compiles and runs fines with x86_64 enabled, that would be default for r16b.  set to default works fine.

camera/texture-view/build.gradle
ggfan: when setting explicitly, application get what required; not to worry from version to version default. In early days, a couple of samples have to be built with C++_* and clang ( gnustl_ won't build ) at that point to align all together, most settings are explicit.

ggfan:
@DanAlbert one more look please: taking as much default as possible:
   1) ndkbuild + gradle: APPSTL need to be set explicitly
   2) gradle experimental need to change to use clang + c++_static as default

will file 2 bugs

audio-echo/app/build.gradle
```diff
@@ -6,22 +6,18 @@ android {
     defaultConfig {
         applicationId 'com.google.sample.echo'
         /*
-           To run on earlier version of Android than android-21, do:
+           To run on earlier version of Android other than android-23, do:
```
DanAlbert: Why is the `minSdkVersion` set so high if it doesn't need to be?

audio-echo/app/build.gradle
```diff
@@ -6,22 +6,18 @@ android {
...
              *) set this minSDKVersion and cmake's ANDROID_PLATFORM to your version
-             *) set ANDROID_STL to c++_static for some very earlier version android.
+             *) set ANDROID_STL to c++_static for earlier Android API levels
```
DanAlbert: Is this a case of `NativeActivity` being broken pre M? There's not enough context in this diff to be very helpful.

audio-echo/app/build.gradle
```diff
@@ -6,22 +6,18 @@ android {
...
-            abiFilters 'armeabi-v7a', 'arm64-v8a', 'x86', 'x86_64'
-        }
+
         externalNativeBuild {
             cmake {
-                // With android studio 2.2.0 +cmake, app needs to pack shared lib to
-                // into APK. Refer to CMakeLists.txt for details
-                arguments '-DANDROID_STL=c++_shared', '-DANDROID_TOOLCHAIN=clang',
+                arguments '-DANDROID_STL=c++_shared',
                           '-DANDROID_PLATFORM=android-23'
```
DanAlbert: Isn't this superfluous?

bitmap-plasma/app/build.gradle
```diff
@@ -7,12 +7,10 @@ android {
...
         minSdkVersion 14
         targetSdkVersion 25
-        ndk {
-            abiFilters 'armeabi', 'armeabi-v7a', 'arm64-v8a', 'x86'
-        }
+
         externalNativeBuild {
             cmake {
-                arguments '-DANDROID_PLATFORM=android-14', '-DANDROID_TOOLCHAIN=clang'
+                arguments '-DANDROID_PLATFORM=android-14'
```
DanAlbert: Superfluous?

audio-echo/app/build.gradle
ggfan: 23 should be the one beginning dynamic run-time permission request.

audio-echo/app/build.gradle
ggfan: no, this does not use the NativeActivity, directly using JNI.  NativeActivity thing is about Android-24 -- pre-N does not load dependency shared lib, N and after NativeActivity does.

For (very )early version of the apps I believe static is flavor recommended. Now it is shared.

audio-echo/app/build.gradle
ggfan: default it is 23, same as minSDKVersion. make it explicit would help for the usage; in some version of NDK, if set to a platform that does not have a android-"version" directory under platforms, it will not build; however minSDK could be set to that version (it exists).

bitmap-plasma/app/build.gradle
ggfan: same as before, mainly sometimes version of minSDK exists, but platforms/android-"$minSDKVersion" does not exist, and build would error out. It is good for application maintenance to explicitly set it: avoiding unnecessary future issues.

ggfan:
the default stl is gnustl_static actually:
```java
if(NOT ANDROID_STL)
  set(ANDROID_STL gnustl_static)
endif()
```
that is inside build/cmake/android.toochain.cmake

the ndkbuild build system actually setting the default to system, I guess that is for the history reasons:
```
# Check that APP_STL is defined. If not, use the default value (system)
# otherwise, check that the name is correct.
APP_STL := $(strip $(APP_STL))
ifndef APP_STL
    APP_STL := system
else
    $(call ndk-stl-check,$(APP_STL))
endif
```
abandon this branch, as it does not fulfil the purpose: default is not c++_static. could not use default in either cmake port or NDK-build. will do it later.

audio-echo/app/build.gradle
DanAlbert: But that's `targetSdkVersion`, not `minSdkVersion`.

audio-echo/app/build.gradle
DanAlbert: > For (very )early version of the apps I believe static is flavor recommended. Now it is shared.

What does API level have to do with this (aside from the bug with `NativeActivity`)? As far as I know the guidance is to use the static STL unless you have multiple shared libraries in your app, in which case you should use the shared STL.

audio-echo/app/build.gradle
DanAlbert: IMO this just clutters up the sample. The fact that this is done in all of the samples (or at least so many of them) makes it look as though the developer must do this rather than rely on `minSdkVersion`. Frankly, there are no for the developer to set this explicitly. Setting this higher than `minSdkVersion` is just broken, and setting it lower needlessly excludes APIs they might want to use.

It would make more sense to me to have *one* sample for all the different configuration options, but keep all the others minimal to make them easier to understand for a newcomer.

bitmap-plasma/app/build.gradle
DanAlbert: This isn't the case with new NDK versions. There are no gaps.

DanAlbert:
> the default stl is gnustl_static actually

Only pre-r16.

That's not relevant to the point I was making though. The point is that the samples should use the NDK defaults when possible. If the sample builds fine with gcc or clang, gnustl or libc++, just use the defaults. If the sample can only be built with clang or only with libc++, set it explicitly (at least until r18, which removes gcc and gnustl/stlport, has been out for a while and the options become obsolete).

For ndk-build, you'll need to set an STL explicitly for anything that uses the STL (I assume most of the samples do, but I don't know that for certain) since we still default to system (essentially no STL). For anyone else coming here, this got filed as https://github.com/android-ndk/ndk/issues/595 and is subject to change.

ggfan:
Sometimes applications might like to make it explicit: certainty -- to build one app, needs certain compiler + stl, no dependency on default tool settings( default settings changing from time to time, and sometimes do not respect backward compatibility nature).  Sometimes it becomes a burden for application maintenance that just because of tool change, many changes have to be done inside source code.  Probably should add a few test to validate toolchain default setting for releases, and leave application settings to application developers.

PR #471, opened 2017-12-13
Remove the deprecating armeabi ABI from build

ggfan:


ggfan:
Purpose:
- remove armeabi
- use default for ABI when possible for samples ( for r16, the default is x86, x86-64, armeabi-v7a, arm64-v8a)
- For ndk-build + gradle, remove explicit clang usage since it is default for r16

will add another branch to deal with CMake's usage for default stl and compiler after fully configuring one or two samples.

@DanAlbert  @jomof please take a look

hello-jni/app/build.gradle
```diff
@@ -58,16 +52,10 @@ android {
...
-            dimension 'cpuArch'
-            ndk {
-                abiFilters 'mips', 'mips64'
-            }
-        }
         universal {
             dimension 'cpuArch'
             ndk {
-                abiFilters 'mips', 'mips64', 'x86', 'x86_64'
+                abiFilters 'x86', 'x86_64'
```
DanAlbert: What's this? How is it universal if it only targets half of the architectures?

other-builds/ndkbuild/hello-neon/app/build.gradle
```diff
@@ -37,10 +37,6 @@ android {
             dimension "abi"
             ndk.abiFilters 'armeabi-v7a'
         }
-        x86 {
```
DanAlbert: I don't speak gradle, but isn't this removing support for x86* from this sample?

hello-jni/app/build.gradle
ggfan: this is just an example to show that application could pick up any kinds of combination of the ABI into flavor build.

other-builds/ndkbuild/hello-neon/app/build.gradle
ggfan: this is a battle: like not to build for x86 as the performance number on x86 is slower than the C version of the sample;  it makes great senses for armeabi-v7a though.

DanAlbert: So? It's a sample. If the NEON emulation performance isn't even good enough for a sample we should just remove it entirely.

hello-jni/app/build.gradle
DanAlbert: I still don't understand how this is in any way "universal", but it's not related to the rest of the patch, so whatever. Should probably add some comments explaining why this is the way it is (I sure don't understand it and as far as gradle knowledge goes, I am the target user of the samples), but that can be a different patch.

other-builds/ndkbuild/hello-neon/app/build.gradle
ggfan: probably the best action now is: keep the x86 build as-is ( the purpose of the PR is to get rid of the warning of armeabi in r16).
created a bug to look at it deeply:  https://github.com/googlesamples/android-ndk/issues/472

hello-jni/app/build.gradle
ggfan: yeah, will look at that with next patch for the compiler and stl default handling.

PR #477, opened 2017-12-27
Update hello-neon README.md to call out release build

ggfan:
This is mainly to clarify https://github.com/googlesamples/android-ndk/issues/474 to use release ( signed APK ) when comparing performance numbers.


ggfan:
Only readme file change, merging it in.

PR #482, opened 2018-01-06
Cleaning up gles3jni sample, details in:

ggfan:
Details in:
    https://github.com/googlesamples/android-ndk/issues/481
The usage of dynamic detection of ES3 is for the case:
  1) want to BUILD APK to support platforms less than API_LEVEL 18; in this case, NDK could only support ES2 header/lib files
  2) the sample APK, when RUNNING on API_LEVEL >= 18 ( for example, Android-27),  could use OpenGL ES3 ( which is for sure available ); but since APK was building with an API_LEVEL that does not have ES3 header/lib, App needs to load the lib and add header file in itself code.

The original comment is confusing, hope this clarifies the things.  The source is shared by 3 build systems( why need 3 ?), so the other 2 build scripts also need to be changed. But this change is good change, I believe.

@ethteck  PTAL: if no objections, I will merge it in, Thanks!


ethteck:
Thanks for clearing everything up! Looks great.

PR #495, opened 2018-02-12
add ndkbuild for nn_sample in master branch

ggfan:
Kind of addressing https://github.com/googlesamples/android-ndk/issues/494, giving NN a special treatment: for new projects we are recommending to use cmake to build.


ggfan:
@mwgray

PR #501, opened 2018-02-23
Set float when requesting FP16 pixel format

courtney-g:
Set color component type to fload for FP16 not 10:10:10:2.
Also set color component type to fixed for all other modes just to
be complete.


ggfan:
Thank you Courtney, you are fast.

PR #506, opened 2018-03-09
Remove mips from san-angeles' ABI list.

rprichard:
The ndk-build version of this sample already has this list of ABIs:

    include 'armeabi-v7a', 'arm64-v8a', 'x86'

Bug: https://github.com/googlesamples/android-ndk/issues/505


ggfan:
LGTM

PR #516, opened 2018-04-26
Update Teapot CMake scripts for review

ggfan:


ggfan:
This is to update CMake scripts regarding to [CMake Video](https://www.youtube.com/watch?v=bsXLMQ6WgIk&t=2s) of https://www.youtube.com/watch?v=bsXLMQ6WgIk&t=2s
@hak @bbilodeau @jomof please take a look, thank you very much!

ggfan:
adding @alexander-a-kharlamov too

teapots/common/ndk_helper/CMakeLists.txt
```diff
@@ -1,16 +1,35 @@
...
+set_target_properties(ndkHelper
+  PROPERTIES
+    CXX_STANDARD 11
+    CXX_STANDARD_REQUIRED YES
+    CXX_EXTENSIONS NO
+)
+target_compile_features(ndkHelper
+  PUBLIC
+    cxx_noexcept
+    cxx_nullptr
```
plabath: What does this do? Is it really necessary. I would expect nullptr to be subsumed by CXX_STANDARD 11

teapots/classic-teapot/src/main/cpp/CMakeLists.txt
```diff
@@ -18,35 +18,52 @@ cmake_minimum_required(VERSION 3.4.1)
...
+    TeapotRenderer.cpp
+)
+set_target_properties(TeapotNativeActivity
+  PROPERTIES
+    CXX_STANDARD 11
+    CXX_STANDARD_REQUIRED YES
+    CXX_EXTENSIONS NO
+)
+target_compile_features(TeapotNativeActivity
+  PRIVATE cxx_noexcept
```
plabath: You're not using the visibility of the cxx_noexcept feature consistently. In some targets you have it as PRIVATE, and in some as PUBLIC. Which one is better is a different debate, but at least you could be consistent about it.

teapots/more-teapots/src/main/cpp/CMakeLists.txt
```diff
@@ -17,36 +17,51 @@
...
+    CXX_STANDARD 11
+    CXX_STANDARD_REQUIRED YES
+    CXX_EXTENSIONS NO
+)
+target_compile_features(MoreTeapotsNativeActivity
+  PRIVATE cxx_noexcept
+)
+target_compile_options(MoreTeapotsNativeActivity
+  PRIVATE
+    -Wall -Werror -fno-rtti
```
plabath: The fact that you have to specify -fno-exceptions and -fno-rtti (which are generally used together) looks weird. If there isn't a compile feature for rtti, then I'd consider specifying -fno-exceptions explicitly, just so these two things can be together.

teapots/more-teapots/src/main/cpp/CMakeLists.txt
```diff
@@ -17,36 +17,51 @@
...
 # build the ndk-helper library
-set(ndk_helper_dir ../../../../common/ndk_helper)
-add_subdirectory(${ndk_helper_dir} ndk_helper)
+set(ndkHelperSrc ../../../../common/ndk_helper)
+add_subdirectory(${ndkHelperSrc} ndkHelperBin)
 
-# Export ANativeActivity_onCreate(),

+# now build app's shared lib
+add_library(MoreTeapotsNativeActivity
```
plabath: The `nativeAppGlue` and `MoreTeapotsNativeActivity` libraries are named inconsistently.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
```diff
@@ -17,36 +17,53 @@
...
+  PRIVATE cxx_noexcept
+)
+target_compile_options(ChoreographerNativeActivity
+  PRIVATE -Wall -Werror -fno-rtti
+)
 
-target_include_directories(ChoreographerNativeActivity PRIVATE
+# Export ANativeActivity_onCreate(),
+# Refer to: https://github.com/android-ndk/ndk/issues/381.
+set(CMAKE_SHARED_LINKER_FLAGS
```
plabath: Why is this not a target property (`set_property(TARGET Foo APPEND_STRING LINK_FLAGS " -u ANativeActivity_onCreate")`)?

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
```diff
@@ -17,36 +17,53 @@
 cmake_minimum_required(VERSION 3.4.1)
 
 # build native_app_glue as a static lib
-add_library(native_app_glue STATIC
-    ${ANDROID_NDK}/sources/android/native_app_glue/android_native_app_glue.c)
-
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11 -Wall -fno-exceptions -fno-rtti")
+add_library(nativeAppGlue STATIC
+  ${ANDROID_NDK}/sources/android/native_app_glue/android_native_app_glue.c)
```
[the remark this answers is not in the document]
gpx1000: Also, it helps when crafting projects built for multiple deployment targets to guard any platform specific code or dependencies.  Thus this could be changed to:
if(ANDROID) //ANDROID is defined in the NDK CMake toolchain
include(AndroidNdkModules)
android_ndk_import_module_native_app_glue()
endif()

Now later when you want to use it just add it to the target like so:
if(ANDROID)
target_link_libraries(ChoreographerNativeActivity native_app_glue)
endif(ANDROID)
By breaking up the target_link_library calls you allow platform specific decisions at build time.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
```diff
@@ -17,36 +17,53 @@
...
+set(CMAKE_SHARED_LINKER_FLAGS
+  "${CMAKE_SHARED_LINKER_FLAGS} -u ANativeActivity_onCreate")
+
+target_include_directories(ChoreographerNativeActivity
+  PRIVATE
     ${ANDROID_NDK}/sources/android/cpufeatures
     ${ANDROID_NDK}/sources/android/native_app_glue
-    ${ndk_helper_dir})
+    ${ndkHelperSrc}
+)
```
gpx1000: Might I suggest here that cpufeatures and native_app_glue include lines are not needed?

The reasoning is CMake will use targets to determine what include directories are required.  i.e. in the target_link_libraries when using native_app_glue and cpufeatures both targets have interface include headers property defined, so CMake is smart enough to allow you to grab the directories for free by simply linking them.
Also, ndkHelperSrc is already included by virtue of the fact that's where the source files are (you only need to provide CMake a custom include directory if it is not the same place as the c++ files are located).
I'd venture this entire target_include_directories is not needed.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
```diff
@@ -17,36 +17,53 @@
...
+add_subdirectory(${ndkHelperSrc} ndkHelperBin)
 
 # now build app's shared lib
-add_library(ChoreographerNativeActivity SHARED
+add_library(ChoreographerNativeActivity
+  SHARED
     ChoreographerNativeActivity.cpp
-    TeapotRenderer.cpp)
+    TeapotRenderer.cpp
+)
```
gpx1000: I like trying to separate out common code from platform specific code.  Might I suggest that TeapotRenderer.cpp can be common to any iOS/Android style targeting at some point in the future; thus structure the lib to have something like this:
add_library(ChoreographerNativeActivity SHARED TeapotRenderer.cpp)
if(ANDROID)
add_library(androidActivityLib STATIC ChoreographerNativeActivity.cpp)
target_link_libraries(ChoreographerNativeActivity ActivityLib)
endif()

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
gpx1000: I believe these are the droids you're looking for:
set_target_properties( ChoreographerNativeActivity PROPERTIES LINK_FLAGS "-u ANativeActivity_onCreate" )

teapots/common/ndk_helper/CMakeLists.txt
```diff
@@ -1,16 +1,33 @@
...
+)
+target_compile_options(NdkHelper
+  PRIVATE
+    -Wall -Werror -fno-rtti
+)
+
+target_include_directories(NdkHelper
+  PRIVATE
+    ${ANDROID_NDK}/sources/android/native_app_glue
+)
```
gpx1000: If you change the above three to PUBLIC from private, then you can eliminate all the other places that rely upon NdkHelper.  Additionally, if you have a common include folder, you can take advantage of interfaces like this:
set_target_properties( NdkHelper PROPERTIES INTERFACE_INCLUDE_DIRECTORIES <commonPath> )

Also, add this:
target_link_libraries(NdkHelper cpufeatures native_app_glue android log)
Now whenever you use NdkHelper in the parent target_link_libraries lines, you don't need any of them.
Finally, native_app_glue doesn't need to be specifically called out here at all; you can use:
include(AndroidNdkModules)
android_ndk_import_module_native_app_glue()
then the native_app_glue target lib is defined.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
```diff
@@ -17,36 +17,53 @@
 cmake_minimum_required(VERSION 3.4.1)
 
 # build native_app_glue as a static lib
-add_library(native_app_glue STATIC
-    ${ANDROID_NDK}/sources/android/native_app_glue/android_native_app_glue.c)
-
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11 -Wall -fno-exceptions -fno-rtti")
+add_library(NativeAppGlue STATIC
+  ${ANDROID_NDK}/sources/android/native_app_glue/android_native_app_glue.c)
```
DanAlbert: Would appreciate keeping the sweeping stylistic changes out of this change since it just makes it harder to review by bloating the diff.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
```diff
@@ -17,36 +17,53 @@
...
-set(ndk_helper_dir ../../../../common/ndk_helper)
-add_subdirectory(${ndk_helper_dir} ndk_helper)
-
-# Export ANativeActivity_onCreate(),

-# Refer to: https://github.com/android-ndk/ndk/issues/381.
-set(CMAKE_SHARED_LINKER_FLAGS
-    "${CMAKE_SHARED_LINKER_FLAGS} -u ANativeActivity_onCreate")
+set(ndkHelperSrc ../../../../common/ndk_helper)
+add_subdirectory(${ndkHelperSrc} ndkHelperBin)
```
gpx1000: A cool trick that helps with large projects:

add_subdirectory(${ndkHelperSrc} ndkHelperBin) can be changed to
add_subdirectory(${ndkHelperSrc} ${CMAKE_CURRENT_BINARY_DIR}/ndkHelperBin)
The difference is subtle but important.
AS will do a different build per variant + per flavor + per build type.  So a good acceleration is to specify that the gradle should build to a folder other than .externalNativeBuild and reuse build bin paths.  This means the trick is to tell cmake that it should create a custom bin path and use that everywhere.  To take advantage of it, our binary obj build path needs to exist under the current bin directory.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
DanAlbert: I think given that the goal of this sample is to show how to use the choreographer API (unless I'm assuming too much based on the directory name), it's best to keep it focused on that. If TeapotRenderer.cpp were generalized and shared among multiple samples that would be a different story, but I think remaining concise is better here.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
DanAlbert: +1. The spirit of the linked talk is *reducing* unnecessary duplication.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
```diff
@@ -17,36 +17,53 @@
...
-    native_app_glue
+    log
     atomic
     EGL
     GLESv2
-    log
-    ndk-helper)
+    NativeAppGlue
+    NdkHelper
+)
```
DanAlbert: Is there any particular reason for changing the ordering here? Link order does actually matter (it needs to be a topological sort of the dependency graph). If NativeAppGlue of NdkHelper use liblog then this is wrong.

teapots/classic-teapot/src/main/cpp/CMakeLists.txt
DanAlbert: `cxx_noexcept` is not the same as `-fno-exceptions`. This is actually saying the opposite: this project requires that the `noexcept` keyword be available, which requires `-fexceptions`.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
ggfan: that is good one as building time is import! one more step:  as this ndkHelper module is supporting 3 projects, each CMake_CURRENT_BINARY_DIR will not be the same, so it will still build 3 times, correct?  What happen if I hardcode to a common directory as bin, then 3 project will be able to use it, instead of building it 3 times, correct?

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
ggfan: yeah, the order is: which one closer to the final module so file, get to the last. More basic a lib is. the fore front it is.  we could adopt the opposite sequence: the more closer to the final so file, the more front it is.  Alphabetic order is ok, but more content aware logical way is better.

Same goes for header file include sequence, although we are not reinforce it too much ( framework guys seems to be good at this )

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
gpx1000: Yep,  If you hardcode it to be a certain path then CMake is smart enough to decide that the output already exists and thus wont need rebuilding.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
```diff
@@ -17,36 +17,53 @@
 cmake_minimum_required(VERSION 3.4.1)
 
 # build native_app_glue as a static lib
-add_library(native_app_glue STATIC
-    ${ANDROID_NDK}/sources/android/native_app_glue/android_native_app_glue.c)
-
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11 -Wall -fno-exceptions -fno-rtti")
```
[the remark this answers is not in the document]
gpx1000: If I might do an addendum here:
if you define the flags (-fno-rtti against a common dependency such as ndkHelperLib) then the parent receives it automatically.  Thus no need for defining it in more than one place.
+1 for removing no-rtti/no-exceptions in a different change.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
gpx1000: More to the point.  Let CMake do your ordering work for you!

That is to say, ndkHelper and NativeAppGlue both might depend upon log.  If both are targets, you can use:
target_link_libraries(ndkHelper log)
target_link_libraries(NativeAppGlue log)
target_link_libraries(ChoreographerNativeActivity nativeAppGlue ndkhelper)

log gets included before either nativeAppGlue and ndkHelper and doesn't get repeated.  The real power of the target like system is to never have to worry about platform ordering.  Cmake will handle it for you.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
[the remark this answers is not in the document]
gpx1000: My first github code review, this comment thread appears collapsed; wanted to ensure this little gem is noticed as it's very useful to be able to avoid hard ndk paths in user land CMakeLists.txt so bump!

teapots/classic-teapot/src/main/cpp/CMakeLists.txt
ggfan: yeah, for the cxx_noexcept, it is client of the compile flag -fno-exception. Will move it into the compile flag and wait for some better ways, thanks!

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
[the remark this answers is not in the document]
ggfan: this is one is done, will be there in the next upload!

ggfan:
All: thanks for review -- it is much cleaner and closer to Daniel's standard! Tried to avoid the source changes except the one for cpufeatures ( it is not used in the source code actually ).  Please have another look, thanks!

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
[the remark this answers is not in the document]
DanAlbert: > if you define the flags (-fno-rtti against a common dependency such as ndkHelperLib) then the parent receives it automatically.

That's an odd behavior imo. Just because a library you use doesn't use RTTI/exceptions doesn't mean that your library shouldn't.

Regardless, it's less intention revealing. Adding it to `CXX_FLAGS` very obviously says "I want these flags applied to every C++ file".

teapots/common/ndk_helper/CMakeLists.txt
```diff
@@ -1,16 +1,31 @@
...
-                           ${ANDROID_NDK}/sources/android/native_app_glue)
+set_target_properties(NdkHelper
+  PROPERTIES
+    CXX_STANDARD 11
+    CXX_STANDARD_REQUIRED YES
+    CXX_EXTENSIONS NO
+)
+target_compile_options(NdkHelper
+  PUBLIC
+    -Wall -Werror -fno-exceptions -fno-rtti
```
DanAlbert: I still think CXX_FLAGS is the right way to do this.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
DanAlbert: > More basic a lib is. the fore front it is.

That's the opposite of what it needs to be. If this is working then the libraries that are having their order shuffled here don't depend on each other, but if that's the case then why are you changing it?

teapots/common/ndk_helper/CMakeLists.txt
ggfan: understood. Since it is norm to go target oriented, we go this "modern" way. It is like global variable vs. wrapped into object. In the end it is same or similar.

ggfan:
@gpx1000 still have 2 things to solve 

- look at ndkHelper's CMakeLists.txt, how to get ${ANDROID_NDK}/sources/android/native_app_glue automatically populated to ndkHelper's client implicitly? is it because ndk's native_app_glue() need to use target_include_directories(PUBLIC...)?
- on command line, I have 2 std=c++11 ( duplicate ). do not know why, but I am going to hack it out.

Please provide the tips here, thanks!

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
```diff
@@ -17,36 +17,53 @@
 cmake_minimum_required(VERSION 3.4.1)
```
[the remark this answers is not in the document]
ggfan: version probably earlier than 3.6.2:  when I built with 3.6.2, it complaint that I only have 3.6.0-rc1.  The first time inside SDK maybe earlier than that, this C side code does not require 3.6.2, it certainly could build ( I could not change to 3.6.2, even I could not build with that version )

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
gpx1000: Admittedly it's probably better to just make an independent crops platform example

teapots/common/ndk_helper/CMakeLists.txt
gpx1000: Sorry writing this on my etcha-sketch.

Cxx_flags is a property so it can obey the target conventions.  Actually can set it in above function and lose this entirely.
A feature of using cxx_flags is it only gets added to c++files.  This is important if a different language is in a project and you want different flags per language.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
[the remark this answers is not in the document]
gpx1000: I'm still on etcha-sketch so can't test this, but this might be not inclusive minimum.  I.e. set it to one less as it expands to error if version is equal to or less than.
I think 3.6.1 is correct here.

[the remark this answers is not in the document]
ggfan: when CMake was supported, the version is 3.6?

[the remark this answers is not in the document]
ggfan: https://b.corp.google.com/issues/37102629
Looks like the first time we support CMake is 3.4.1.   I would like to stay as is so different Android Studio version could still compile the code.  Thx

teapots/common/ndk_helper/CMakeLists.txt
ggfan: that is good to know.  I believe this project is all CPP files, so they could apply to all.
It might be good to show some variety of usage.  thx

ggfan:
Thanks for review!  Getting close to the end of it, still need one thing to populate header files from dependent modules. The other one is solved ( C++ stl adds c++11 automatically ), no issue.

gpx1000:
Will try to look closer at it tonight from hotel.  Am writing this from side of mountain limited cell signal.

You should be able to use public or interface to tell all dependent modules.  Where is the issue and might be able to tell answer from here.  Or can wait for tonight when have WiFi.

gpx1000:
I haven't been able to pull out laptop yet.  But here's the way I like to explain this strategy as I think this might be the issue.
Each target is analogous to an object.  It has visibility necessary only to build the object itself it doesn't automatically propagate up things that are set to private within it's members.
So if the function include-native app glue is set to private which I think it is against the target then it won't auto propagate up.
To get around this, set the property interface include headers against the ndk helper target to get generated from the native app glue interface property.
Doing that will allow any consuming target to get the includes automatically.

teapots/common/ndk_helper/CMakeLists.txt
plabath: I agree with Dan. If your intention is for something to apply to the entire project, then it should be specified as a global, instead of relying on the implicit propagation of options through dependencies.

teapots/classic-teapot/src/main/cpp/CMakeLists.txt
```diff
@@ -16,37 +16,32 @@
...
-# build the ndk-helper library
-set(ndk_helper_dir ../../../../common/ndk_helper)
-add_subdirectory(${ndk_helper_dir} ndk_helper)
-
-# Export ANativeActivity_onCreate(),

-# Refer to: https://github.com/android-ndk/ndk/issues/381.
-set(CMAKE_SHARED_LINKER_FLAGS
-    "${CMAKE_SHARED_LINKER_FLAGS} -u ANativeActivity_onCreate")
+# build the ndk-helper library into a common directory
```
plabath: This looks like a hack to me, and I'm not sure we should be encouraging it. It violates layering by encoding knowledge about other projects here, even though this should be a standalone project as far as cmake is concerned. If there are things that are to be shared between multiple modules/projects, then I think that should be handled at a higher level (either a separate module/project in gradle, or something else).

ggfan: the new version is changed, thanks!

ggfan:
@gpx1000 and others, here is the situation and I would think there is a decent solution ( it is quite common case ):

- C/C++ side TeapotNativeActivity depends on both ndkHelper  & NativeAppGlue
- ndkHelper depends on NativeAppGlue

NativeAppGlue's header file is configured with include_directories:
```
  include_directories(${ANDROID_NDK}/sources/android/native_app_glue)
```
so its direct clients could use it. But this include_directories() clause does not seem to populate up through direct clients to other indirect higher layer clients, as in ndkHelper, it is:
```
target_include_directories(NdkHelper
  PUBLIC
    ${CMAKE_CURRENT_SOURCE_DIR}
)
```
TeapotNativeAcivity could not find the native_app_glue header file. It could be added by add:
```
    ${ANDROID_NDK}/sources/android/native_app_glue
```
into NdkHelper's target_include_directories().  But what is the clause to populate at target level directly without hardcoding it: force native_app_glue header file to be exposed in NdkHelper target (to NdkHelper's clients) WITHOUT changing SDK's implementation of native_app_glue() function. 

Or I use it wrong? Thanks

gpx1000:
So amazingly as soon as I sent the last message about generation, I forgot that it could be useful here too.  Try this:

set_target_properties(NdkHelper
  PROPERTIES
    CXX_STANDARD 11
    CXX_STANDARD_REQUIRED YES
    CXX_EXTENSIONS NO
    INTERFACE_INCLUDE_DIRECTORIES $<TARGET_PROPERTY:NdkHelper,INCLUDE_DIRECTORIES>
)

That should capture the idea of force all include directories NdkHelper knows about it to be what is consumed by everyone above it.  It's a kludge and really this would be cleaner:
set_target_properties(native_app_glue $<TARGET_PROPERTY:NdkHelper,INCLUDE_DIRECTORIES>)
but include_directories makes this a not clean solution all around unfortunately.

ggfan:
@gpx1000 Thanks for the comment!
"The other option is simply to place in each project a call to android_ndk_import_module_native_app_glue()
That way each project gets a include_directories() call."
this one does not work as CMake does not allow for add_library() for the same module more than once :-(.
So my option is your 1st or my-own old-fashioned way of CMake,  you are the winner here: first approach is the one we should go and will be going :-(! I will update it later today and let you know!

As for SDK's implementation, will follow-up to get it changed: but our change should not depend on that future change ( lot of users are in the same shoes as I am now, and they need a solution )

teapots/classic-teapot/src/main/cpp/CMakeLists.txt
jomof: I agree Android Studio shouldn't be involved in the factoring since it is a native-to-native dependency. I don't think @ggfan has any options that are better than add_directory. And, assuming there's a reason ndk_helper is shared as source rather than as .a or .so, I don't know of a better way than add_directory here even with cooperation from the NDK. @DanAlbert jic

DanAlbert: Surely there's a middle ground between this and an AAR. Is there no way to share components between multiple cmake projects in gradle?

ggfan: Yeah CDep could solve the problem: because it has header in the interface, aar is not really a solution for this.

I would like to keep this code change to be CMake centric: not adding another component into the mix. Like to prove that normal CMake should be able to get the work done ( CMake should be enough there is nothing magic here at all ).

gpx1000: Just for completeness sake (recommend against this as it'll confuse lots of people):
https://cmake.org/cmake/help/v3.2/module/CPack.html
CPack technically would allow one to author a package using just cmake style stuff specifying the "install" location.  Then a consuming level app built separately with gradle module, can consume it by find_package() syntax.
Thus there does exist a solution that would be very similar to a cdep style solution.  However, getting it right is tricky to the point of "I personally haven't gotten it to work right."  I do use this style trick with the PS4 variant so I know it works, but cmake add_directory is much cleaner.

teapots/common/ndk_helper/CMakeLists.txt
ggfan: @gpx1000 CXX_FLAGS is not a target_property, correct? I got trouble when trying to use it. and [this place](https://stackoverflow.com/questions/28282896/pass-cmake-cxx-flags-to-target-compile-options) seems to suggest that 2 ways to get it done: set() and target_compile_options(); set() method needs to use appending like "${CMAKE_CXX_FLAGS} then-the-new-ones". it is the old way, we should stick with new way as that link said.

ggfan:
@gpx1000 please take a final look: it is close to the final version now:
1. target_link_library`(NdkHelper, PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})` is still needed, as `INCLUDE_DIRECTIRES` needs someone to put that dir into it. But it does save us from hack of app_glue header file path
2. target_compile_options() is needed for "-Wall -Werror -fno-exceptions -fno-rtti" because CXX_FLAGS is not target property; even we have CXX_FLAGS, we have to write it similar as
```
set_target_properties(NdkHelper, PROPERTIES
   CXX_FLAGS  $<TARGET_PROPERTIES:NdkHelper, CXX_FLAGS> -Wall ... -fno-rtti"
``` 
because setting is NOT appending; and we just want to append -- gradle / NDK add their own flags already.
3. I could NOT remove set_target_properties() inside application's CMakeLists.txt, surprisingly! Probably it is our cmake implementation bug somewhere: the "-std" option ONLY up populated into THE FISRT app in the Android Studio Workspace, the rest of the projects in the same workspace is NOT auto populated with NdkHelper's -std requirement. Strange, but before bottoming that one out, I could not delete the following from application's CMakeLists.txt:
```
set_target_properties(ChoreographerNativeActivity
  PROPERTIES
    CXX_STANDARD 11
    CXX_STANDARD_REQUIRED YES
    CXX_EXTENSIONS NO
)
```
thanks and I like your confirmation for #2 above :-)!

gpx1000:
You're right, my fault.  CMAKE_CXX_FLAGS is a kind of lazy initialization that only works in the cmakelists.txt where it's defined.  So one can do it differently per directory or set it project wide (it is afterall one of the command line variables that can be passed).
Well as in most things, there is another way to skin a cat:

$<$<COMPILE_LANGUAGE:CXX>:${BUILD_FLAGS_FOR_CXX}>

That method would allow you to specify per compile language what flags you want.  You could also target per generator but we're always dealing with ninja/make so not really useful without a cross platform requirement.
The cleaner option that avoids a generator command would be what you already have (target_compile_options).
This led me to a bug in the ninja generator:
set_target_properties(NdkHelper
  PROPERTIES
    COMPILE_FLAGS $<$<COMPILE_LANGUAGE:CXX>:${BUILD_FLAGS_FOR_CXX}>
)
That will cause the ninja generator to incorrectly say: 
bad $-escape (literal $ must be written as $$)
which is this bug: 
http://cmake.3232098.n2.nabble.com/Error-with-ninja-generator-when-using-target-link-library-with-the-debug-keyword-td7594332.html

I think the right solution for now is to go with the add_compile_options and accept that it will not differentiate between c++ files and non-c++ files.  It would be nice to get the other bug fixed as this could be a common use case:
COMPILE_FLAGS $<$<COMPILE_LANGUAGE:CUDA>:${BUILD_FLAGS_FOR_CUDA}>

However, that can be a future thing.

In anycase, see if this is happy for you in order to remove the CMAKE_CURRENT_SOURCE_DIR include line:
set_target_properties(NdkHelper
  PROPERTIES
    CXX_STANDARD 11
    CXX_STANDARD_REQUIRED YES
    CXX_EXTENSIONS NO
    INTERFACE_INCLUDE_DIRECTORIES "$<TARGET_PROPERTY:NdkHelper,INCLUDE_DIRECTORIES> ${CMAKE_CURRENT_SOURCE_DIR}"
)

ggfan:
Oh, thanks for the info! let me test out one case:  whether interface_include_directories are all inside private include directory ( that directory is needed to build ndkHelper module also ). If every directory inside interface_include_directories is automatically included into private, it will work; otherwise we still need both private and interface ( which is public ). I will update

teapots/classic-teapot/src/main/cpp/CMakeLists.txt
DanAlbert: cdep is *an* answer here, but that's really meant as a distribution channel. I wouldn't expect people to use it in a case like this where the dependency is expected to be built as a part of the project (though if ndk-helper were split out into its own project and a cdep package were maintained that was imported here, then I agree).

I was expecting something more along the lines of there being a CMake project for ndk-helper that gets built by gradle into its own project and then there's a FindNdkHelper.cmake that sets up the library and include paths. aiui that's the preferred way of importing externally built libraries into CMake.

teapots/common/ndk_helper/CMakeLists.txt
```diff
@@ -1,16 +1,44 @@
...
+set_target_properties(NdkHelper
+  PROPERTIES
+    CXX_STANDARD 11
+    CXX_STANDARD_REQUIRED YES
+    CXX_EXTENSIONS NO
+    INTERFACE_INCLUDE_DIRECTORIES $<TARGET_PROPERTY:NdkHelper,INCLUDE_DIRECTORIES>
+)
+target_compile_options(NdkHelper
+  PUBLIC
+    -Wall -Werror -fno-exceptions -fno-rtti
```
DanAlbert: To be crystal clear: I will not be approving this change until either this goes or I hear an argument for why this is an improvement beyond "this slide deck says so".

None of these flags should be exported from this module because they are not requirements for using the module. Doing this forces these flags for all *users* of this library despite the fact that they are only relevant for *building* this library. These are policy decisions that should be made either by each library or by the builder for the project as a whole (which is what `CMAKE_CXX_FLAGS` is for).

ggfan:
the syntax:
```
INTERFACE_INCLUDE_DIRECTORIES "$<TARGET_PROPERTY:NdkHelper,INCLUDE_DIRECTORIES> ${CMAKE_CURRENT_SOURCE_DIR}"
```
causing trouble: causing the previously exported native_glue_app header files not visible anymore.  Maybe syntax is wrong ( or a bug in our system )?

teapots/common/ndk_helper/CMakeLists.txt
ggfan: Yeah, I agree that this is module level and this could be done as PRIVATE.  so it limits to this module.  Client could do their own settings for the similar stuff but with different settings.

I am ok to change it to private, and have project to have the similar settings to avoid that global CMAKE_CXX_FLAGS setting, how is that?

no-exceptions and no-rtti was "have to be " used in early days (as I was told) for avoiding build issues and code size ( I think ) so I am not really ready to get rid of them here.

teapots/classic-teapot/src/main/cpp/CMakeLists.txt
gpx1000: I actually am working on using cdep for exactly this kind of purpose (gradle plugin for authoring packages etc).  But that's a future thing.
What you're referencing is actually possible; gradle build that invokes cmake project for ndk-helper into its' own project.  What it'd take isn't what I'd call normal procedures, but here's a quick stab off the top of my head:
from module level gradle:
applicationVariants.all { variant ->
       variant.mergeAssets.doLast {
                copy { 
                   from: <there's a clever closure here involving variant and output files I don't know off top of my head>
                   to: "src/main/jniLibs"
                   include: *.so *.a
               }
            }
        }
 
Then either create a file called FindNdkHelper.cmake that looks like this:
set (LIBNDKHELPER_FOUND true)
set (LIBNDKHELPER_LIBRARIES ${CMAKE_CURRENT_BINARY_DIR}/ndkHelper)
set (LIBNDKHELPER_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/src/include)

Or one could set CMAKE_LIBRARY_PATH to be the output folder where ndkHelper gets built sending that to cmake either through the command line or through a set() command:
https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/Useful-Variables

One could also have the original gradle invocation issue an "install" command to cmake's build process (either through a gradle custom command or through the gradle android studio plugin).
That would be the path of using cpack.

Here's an example for how I solved a somewhat similar problem (have gradle get cmake to output an xcode framework): https://github.com/gpx1000/ARKitARCoreBridge

I'm going to stay out of policy discussions, just here to offer help ;-).  I might suggest however, that if the flags in this case aren't necessary for the NdkHelper object, then they probably do belong in the CMAKE_CXX_FLAGS at the individual project level.  Let me know if I can be of more help (not wanting to get into any debates).

gpx1000:
Hrm... that's what happens when I try to answer off top of head.  I tried a few things, this "should work" but doesn't for me when I test it:
set_target_properties(NdkHelper
  PROPERTIES
    CXX_STANDARD 11
    CXX_STANDARD_REQUIRED YES
    CXX_EXTENSIONS NO
    INTERFACE_INCLUDE_DIRECTORIES $<TARGET_PROPERTY:NdkHelper,INCLUDE_DIRECTORIES>
    INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_CURRENT_SOURCE_DIR}
)
So I think still need target_include_directories.  Sorry for the bug.

ggfan:
Thank you @gpx1000.  Now we are on the same page, that is great ( I think some might depend on our implementation for Gradle, need significant amount of time to see how desktop works then compare, while that is needed, for now we settle down with what we could use :-) )! move on to address others (functionality wise we are good at least for now :-) )

ggfan:
updated to use CMAKE_CXX_FLAGS across board. thanks and it is ready to go for this PR.!

great appreciation to @gpx1000 for your deep knowledge and super helpful tips: pretty sure your efforts will be more appreciated by developers coming to the site for CMake scripts! Thanks again!

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
```diff
@@ -15,38 +15,39 @@
...
+    CXX_EXTENSIONS NO
+)
+
+# Export ANativeActivity_onCreate(),
+# Refer to: https://github.com/android-ndk/ndk/issues/381.
+set_target_properties(${PROJECT_NAME}
+  PROPERTIES  LINK_FLAGS "-u ANativeActivity_onCreate")
 
 # add lib dependencies
-target_link_libraries(ChoreographerNativeActivity
```
DanAlbert: Aren't these needed by this project? (same goes for all the other cases where this changed)

teapots/classic-teapot/src/main/cpp/CMakeLists.txt
DanAlbert: Yeah, something along those lines is more what I would expect, but I was hoping there might be some out of the box support for doing it from gradle. Might be best to back out this part of the change (it's not really related to the rest of the commit anyway) and explore this further until we find the right solution.

teapots/common/ndk_helper/CMakeLists.txt
```diff
@@ -1,16 +1,40 @@
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11 -Wall -fno-exceptions -fno-rtti")
```
DanAlbert: Looks like you dropped this.

teapots/common/ndk_helper/CMakeLists.txt
```diff
@@ -1,16 +1,40 @@
...
-            GLContext.cpp
-            interpolator.cpp
-            JNIHelper.cpp
-            perfMonitor.cpp
-            sensorManager.cpp
-            shader.cpp
-            tapCamera.cpp
-            vecmath.cpp)
+include(AndroidNdkModules)
+android_ndk_import_module_native_app_glue()
```
DanAlbert: Where does this come from? grep doesn't find itself anywhere in this commit or the master branch.

teapots/choreographer-30fps/src/main/cpp/CMakeLists.txt
ggfan: [project](https://cmake.org/cmake/help/v3.0/command/project.html)(ChoreographerNativeActivity LANGUAGES C CXX)
set ${PROJET_NAME} to be the first parameter, and "The top-level CMakeLists.txt file for a project must contain a literal, direct call to the project() command; loading one through the include() command is not sufficient. If no such call exists CMake will implicitly add one to the top that enables the default languages (C and CXX)."

DanAlbert:
Why was this merged? There are still outstanding review comments. It looks like you dropped some necessary code.

teapots/classic-teapot/src/main/cpp/CMakeLists.txt
DanAlbert: Maybe something like this? https://gitlab.kitware.com/cmake/community/wikis/doc/tutorials/Exporting-and-Importing-Targets

teapots/common/ndk_helper/CMakeLists.txt
ggfan: really, let me check, it is strange to drop this

ggfan: it is there, just moved it to top of the file, right after project() statement.

teapots/classic-teapot/src/main/cpp/CMakeLists.txt
ggfan: this is built it into a third party lib, and then import it as an external lib. Which essentially is what cdep is ( and CDep goes one step further: release the binary, not source so no need to build at all)

ggfan:
The code is there. nothing dropped.  Following on issues could file as bugs and address them accordingly, not an issue.

teapots/classic-teapot/src/main/cpp/CMakeLists.txt
gpx1000: @DanAlbert That is an awesome idea.  However, I'm noticing that there's a requirement there of an "install" phase for CMake.  FWIW as far as I know, Android Studio's gradle plugin doesn't utilize the cmake install phase; only build.

However, the interesting thing about this as an idea is if, in the future, there were to be a cmake install phase available, then we can use CPack or any form of exporting/importing targets or even targeting non-android builds.  It would eliminate many of the requirements from other systems.  It also would mean that we could craft a CPack CDep package generator and eliminate the need for a separate gradle publishing plugin.

PR #518, opened 2018-05-02
Removed native-activity & endless-tunnel from Java sample list

ggfan:
Related bug: b2/78915221


proppy:
are there other sample that are native-activity based?

Maybe https://github.com/googlesamples/android-ndk/blob/master/native-plasma ?

ggfan:
seems there are 2 more.

PR #521, opened 2018-05-03
fix for b2/79184992

ggfan:
1. remove  ndk_helper::SensorManager::Process()
2.  add -Wno-attributes for gcc case
When using gcc, 2 warning coming up. Surprisingly that clang does not detect the first one above.  The second one is related to [known issue](https://stackoverflow.com/questions/39830538/ndk-r13-fexceptions-wall-warning-conflicts-with-previous-declaration).

@hak @DanAlbert please carefully review this PR, thanks!


DanAlbert:
IMO better to just stop supporting GCC in the samples. We don't support GCC and this sends the mixed message that we do.

ggfan:
@DanAlbert make sense. I will follow up to test matrix so "this suppress warning" added into their procedure if they want to test gcc using the sample, that way get sample out of trouble.

ggfan:
changed concerning the "suppress warning" part: not adding it for gcc.

PR #545, opened 2018-07-14
Gradle experimental

patrickschneider23:


ggfan:
thank you for your PR.  Could you diff with master-experimental branch? experimental plugin is not advancing since Android Studio 2.3.x, probably put efforts with CMake or ndk-build.  Thanks!

PR #553, opened 2018-08-09
Clone 1.0.0 branch of libwebp instead of master

jhasse:
The master branch fails to build for me atm. In general I think it's a good idea to stick to a specific version, as the master branch might not be stable enough and to make the build reproducible.


jhasse:
I signed it!

ggfan:
Thanks!

Issue #314, opened 2016-10-21
About Teapot Sample


NativeCodeSamples:
Can you provide a Teapot sample that we can learn below things.
1. How to use textures with 3D Teapot model with a use of "bitmap" and ".tga" file ? 

i mean that how to bind textures with a use of .bmp  or .tga files to 3D Model via NDK code.

ggfan:
adding @hak

ggfan:
less than 2 years, got chance to address this one  with https://github.com/googlesamples/android-ndk/pull/555

Loading file from assets probably is the only one specific to Android; this one avoids the reading from disk because I like to avoid to get too much Java code into it ( Run time permission is only available from Java ).

PR #555, opened 2018-08-17
Adding textured-teapot sample for github issue #314

ggfan:
This is to address https://github.com/googlesamples/android-ndk/issues/314, please check for README.md under teapots/textured-teapot for some info, and more in source files.

Will add cubemap when get chance. Textued on Android P and O on pixel2


ggfan:
@owahltinez please take a look, thank you so much!

PR #573, opened 2018-09-13
NativeMedia sample: use application's assetManager

ggfan:
This is to fix issue https://github.com/googlesamples/android-ndk/issues/571:
 should use application's assetManager, not Activity's assetManager to avoid being garbage collected

### DIFF
--- a/native-media/app/src/main/java/com/example/nativemedia/NativeMedia.java
+++ b/native-media/app/src/main/java/com/example/nativemedia/NativeMedia.java
@@ -68,7 +68,7 @@ public class NativeMedia extends Activity {
         setContentView(R.layout.main);
 
         // get application's assetManager to avoid being garbage-collected.
-        assetMgr = getAssets();
+        assetMgr = getApplication().getAssets();
 
         mGLView1 = (MyGLSurfaceView) findViewById(R.id.glsurfaceview1);
         mGLView2 = (MyGLSurfaceView) findViewById(R.id.glsurfaceview2);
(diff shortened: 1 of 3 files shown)


ggfan:
@adamlesinski, please take a look, thank you so much!

adamlesinski:
This doesn't quite solve it. You are still accessing the same AssetManager, but more importantly, a reference to the AssetManager isn't being taken.

I think what really needs to change is android_fopen_set_asset_manager. That method should change to be something like:

AAssetManager* android_asset_manager = NULL;
jobject android_java_asset_manager = NULL;
void android_fopen_set_asset_manager(JNIEnv* env, jobject assetManager) {
   android_java_asset_manager = (*env)->NewGlobalRef(env, assetManager);
   android_asset_manager = AAssetManager_fromJava(env, android_java_asset_manager);
}

native-media/app/src/main/java/com/example/nativemedia/NativeMedia.java
```diff
@@ -67,7 +67,8 @@ public void onCreate(Bundle icicle) {
         super.onCreate(icicle);
         setContentView(R.layout.main);
 
-        assetMgr = getResources().getAssets();
+        // get application's assetManager to avoid being garbage-collected.
+        assetMgr = getAssets();
```
adamlesinski: getAssets() is internally just getResources().getAssets().

This should be getApplication().getAssets().

adamlesinski:
It would be worthwhile, yes. I commented on the commit.

PR #590, opened 2018-10-23
Fix logic to push wrong buffer to free queue

koseonjae:


koseonjae:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here (e.g. `I signed it!`) and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google.com/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).

I signed it. thanks

audio-echo/app/src/main/cpp/audio_player.cpp
```diff
@@ -179,14 +179,18 @@ AudioPlayer::~AudioPlayer() {
...
+      freeQueue_->push(buf);
+    }
   }
   delete devShadowQueue_;
 
   while (playQueue_->front(&buf)) {
     buf->size_ = 0;
     playQueue_->pop();
-    freeQueue_->push(buf);
+    if(buf != &silentBuf_) {
```
ggfan: silent buffer probably never get into the "player queue": only recorder fill the player queue for the recorded audio. 

This one may not be necessary, may you help double check it?  thanks

audio-echo/app/src/main/cpp/audio_player.cpp
```diff
@@ -179,14 +179,18 @@ AudioPlayer::~AudioPlayer() {
   while (devShadowQueue_->front(&buf)) {
     buf->size_ = 0;
     devShadowQueue_->pop();
-    freeQueue_->push(buf);
+    if(buf != &silentBuf_) {
```
ggfan: that is correct: thank you for the fix.

koseonjae: I deleted player queue silent buffer check condition. please check. thanks!

einsteinarbert:
freeQueue_: how can I save it to PCM file by convert to bytes array in java? Does freeQueue_ is the data of recorder bytes?

ggfan:
Yeah, it is byte array ( raw pcm data ):

- recorder pass pcm to app ( us )
- app puts it into reQueue_ (; then get an empty buffer from freeQueue_ and give to recorder)
- player pop a recorded audio buf from playQueue_: playQueue and recQueue are the same queue, shared between player and recorder.  ( player puts the just played buf onto freeQueue -- so recorder could use ).

The PCM format used is controlled by app: frequency, float/int, 16bit/24bit etc.

we are moving on to use [Oboe](https://github.com/google/oboe), please develop with Oboe, same concept but much simpler.

einsteinarbert:
Thanks for your answers, I resolved my issue following
https://github.com/koseonjae/Karaoke

P/S thanks Koseonjae too.

>

einsteinarbert:
Still error, in Oboe, I can't see example for echo app (like Karaoke). Therefore I still using audio-echo example with fixed code from: https://github.com/koseonjae/Karaoke:
```
bool EngineService(void *ctx, uint32_t msg, void *data) {
  assert(ctx == &engine);
  switch (msg) {
    case ENGINE_SERVICE_MSG_RETRIEVE_DUMP_BUFS: {
      *(static_cast<uint32_t *>(data)) = dbgEngineGetBufCount();
      break;
    }
    case ENGINE_SERVICE_MSG_RECORDED_AUDIO_AVAILABLE: {
      // adding audio delay effect
      sample_buf *buf = static_cast<sample_buf *>(data);
      assert(engine.fastPathFramesPerBuf_ ==
             buf->size_ / engine.sampleChannels_ / (engine.bitsPerSample_ / 8));
      engine.delayEffect_->process(reinterpret_cast<int16_t *>(buf->buf_),
                                   engine.fastPathFramesPerBuf_);

      // TODO error crash in some device that has low speed sdcard -> using vector for save data
      /*if(fos == nullptr)
        fos = fopen("/sdcard/recorded_audio.pcm", "wb");
      fwrite(buf->buf_, 1, buf->size_, fos);*/


      sample_buf output = sample_buf();
      output.buf_ = new uint8_t();
      *output.buf_ = *buf->buf_;
      output.size_ = buf->size_;
      bucket.push_back(output);



      break;
    }
    default:
      assert(false);
      return false;
  }

  return true;
}
```
In stop function **stopPlay(JNIEnv *env, jclass type)** - I added:
```
if(fos == nullptr)
        fos = fopen("/sdcard/recorded_audio.pcm", "wb");
  LOGE("====stopPlay %u", bucket.size());
  for(auto const& value: bucket) {
    fwrite(value.buf_, 1, value.size_, fos);
  }
  fclose(fos);
  fos = nullptr;
```
I don't know why the fwrite which write directly to sdcard work (but some crash in low speed sdcard devices) in EngineService, My Vector bucket write out ok but cannot hear anything. 
Does someone explane it? Something wrong here?
P/S: I think write to memory like vector is faster than sdcard? It true, isn't it? But not working

ggfan:
EngineService is inside Audio recorder's callback thread, could not do file writing: file access is blocking I/O, there just no time to do this in audio thread:
   - while waiting for writing to happen, player would already running out buffer to play
   - recording side would over-run big time.
Disk access could not be inside audio thread.  Copying out to memory seems ok to me ( audio data is not that big, comparing to video ), could not directly write to disk from audio thread. I am not familiar with the implementation of the framework: there got to be a queue between disk-writing and incoming audio thread.

einsteinarbert:
so, how to disable/enable echo mode?

ggfan:
Playing silent audio but keeps the streaming going is one way of stopping, and resuming playing back real recorded audio would be resume it. This way app might be more smooth ( escaped the so called "audio warm-up" period ) and app having more control. The downside is that it is always running (but might have a better audio experience overall ) 

Or tearing down everything to stop, re-starting from scratch for starting. This way app might experience glitch audio; and starting / stopping might has complication ( as recording and playing back are in the different thread, make sure releasing resources AFTER no callback threads are still running )

einsteinarbert:
Thanks for reply, but if I we can modify volume level of micro to zero? I tried it but I can't find where to do in audio-echo source code :(

ggfan:
not in the source code, check for  https://www.khronos.org/registry/OpenSL-ES/api/1.1/OpenSLES.h, for 
  Volume Interface / Mute

you would add them in.

PR #624, opened 2019-03-01
Android-ndk

Bhanditz:
Ndk build


ggfan:
@Bhanditz, may you follow this CLA requirement:
 **Please visit https://cla.developers.google.com/ to sign**

thanks!

ggfan:
Thank you for your patch.

The new samples are added in master branch ( CMake ); ndkbuild is supported from [other-build/ndkbuild](https://github.com/googlesamples/android-ndk/tree/master/other-builds/ndkbuild) in the master branch in the favor of source code maintenance:  bugfix just needs to check into one place, not the many places.

Thank you for the efforts to bring up master-ndkbuild to match master branch. Due to the above the mentioned reasons, it might take some efforts to get it done.  You could break it down to smaller CLs to make things manageable. The .ci_tools directory probably needs some serious validation before merge in, like to recommend to leave it out for now.

Thank you again for your efforts!

ggfan:
this might be a mistake, closing it down, please break it into small changes then we could have a look

PR #638, opened 2019-05-02
Remove invalid call to glHint()

### LINKED ISSUE #637: Invalid call to glHint()
https://github.com/googlesamples/android-ndk/blob/86ceeb248bb20bfffb09330d1a95bcf3e91bb99d/native-activity/app/src/main/cpp/main.cpp#L149

This glHint() call is invalid because GL_PERSPECTIVE_CORRECTION_HINT is not listed in the valid parameter in GLES 3 specification, see https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glHint.xhtml.

hevrard:
Parameter GL_PERSPECTIVE_CORRECTION_HINT is not among the valid ones
with respect to the GLES 3 specification, see:
https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glHint.xhtml

fix #637

### DIFF
--- a/native-activity/app/src/main/cpp/main.cpp
+++ b/native-activity/app/src/main/cpp/main.cpp
@@ -146,7 +146,6 @@ static int engine_init_display(struct engine* engine) {
         LOGI("OpenGL Info: %s", info);
     }
     // Initialize GL state.
-    glHint(GL_PERSPECTIVE_CORRECTION_HINT, GL_FASTEST);
     glEnable(GL_CULL_FACE);
     glShadeModel(GL_SMOOTH);
     glDisable(GL_DEPTH_TEST);


hevrard:
This app actually tragets GLES 1.0 apparently, in which case this call is valid.
Should we update this sample to use GLES 3.2 ?

ggfan:
we probably should enable this one as GLES 3.1 or create a new one to use. 
A little more work to update it:
  - need to use right headers as GL headers are in GLESx/  directory ( x: "nothing", 2, 3 )
  - pull in the right lib

ggfan:
re-thinking about this one:  there are 2 other ones for gl2/gl3 purpose, leaning to keep this one as-is ( for 1.0 ).

PR #648, opened 2019-08-12
Minor update to hello-oboe sample concerning error handling.

ggfan:


hello-oboe/app/src/main/cpp/hello-oboe.cpp
```diff
@@ -23,11 +23,17 @@ static OboeSinePlayer *oboePlayer = nullptr;
 
 
 extern "C" {
-    JNIEXPORT void JNICALL
+    /* Create Oboe playback stream
+     * Returns:  0 - success
+     *          -1 - failed
+     */
+    JNIEXPORT jint JNICALL
```
nic0lette: This doesn't feel like idiomatic return values for Java. It's fine for C++, but (imo) it should be returning a jboolean set to 'true' for success and 'false' for failure. Then the code below is:

```
if (createStream()) {
}
```

Which feels much more natural to me.

hello-oboe/app/src/main/java/com/google/example/hellooboe/MainActivity.kt
```diff
@@ -21,6 +21,8 @@ import androidx.appcompat.app.AppCompatActivity
 import android.os.Bundle
 import android.view.MotionEvent
 import android.view.View
+import android.widget.Toast
+import kotlinx.android.synthetic.main.activity_main.*
```
nic0lette: Don't use the synthetics, or at least don't import * with them. :)

hello-oboe/app/src/main/java/com/google/example/hellooboe/MainActivity.kt
```diff
@@ -44,7 +46,12 @@ class MainActivity : AppCompatActivity() {
 
     override fun onResume() {
         super.onResume()
-        createStream()
+        if (createStream() != 0) {
+            val errorString : String = getString(R.string.error_msg)
+//            "Error: unable to create Oboe stream \n please report issue @ https://github.com/google/oboe"
```
nic0lette: nit: You can remove this comment, I think?

hello-oboe/app/src/main/cpp/hello-oboe.cpp
ggfan: agreed and good point.  It is the point that we discussed and finally agreed upon for audio that we will return int instead of bool in favor of future compatibility ( we might want to populate future more error code types to Java side ).

ggfan:
@philburk PTAL

hello-oboe/app/src/main/java/com/google/example/hellooboe/MainActivity.kt
ggfan: good catch: done!  thx

hello-oboe/app/src/main/java/com/google/example/hellooboe/MainActivity.kt
ggfan: done, thanks!

PR #657, opened 2019-10-20
fix internal bug b/142666190

ggfan:
NDK R21 fixed gnu make's abspath issue on Windows Host, no need to use
the workaround. After NDK R25 is released, will totally remove
abspath_wa() in Android.mk files

Will merge after NDK-R21 is formally released


ggfan:
the other branch is merged and this one is abandoned.

PR #658, opened 2019-10-26
Improve libraries build code

joe-skb7:
Changes:
  1. Bring back strong dependency from "gen-libs" module in "app" module (workaround is not needed anymore)
  2. Make it possible to do repeated (incremental) build, when libraries are build as well. It might be not so interesting in the context of this sample project, but it is very helpful in real projects which build native libraries along with Java code, rather than using pre-built lib versions. This is usually the case when such library is only being developed for some particular app.

After updating to new Gradle plugin, the "wrong format" issue is reproducing with 100% probability, that's why I decided to send those patches (which help in my pet project), and that's why I included Gradle plugin update patch in this patch series.

See commit messages for details.

### DIFF
--- a/hello-libs/app/build.gradle
+++ b/hello-libs/app/build.gradle
@@ -38,20 +38,8 @@ android {
 
 dependencies {
     implementation 'com.android.support:appcompat-v7:28.0.0'
+    // uncomment out this one to generate lib binaries,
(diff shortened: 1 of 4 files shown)


ggfan:
thanks for the PR! merged!

PR #665, opened 2019-11-14
Drop bug in workaround for r21+.

DanAlbert:
This workaround for a bug in abspath used to workaround path length
issues in Windows is not needed in r21+ because the abspath bug was
fixed. Unfortunately the bug being fixed means that the workaround now
introduces a bug when forming the path. Only use the workaround for
pre-r21 NDKs.

Bug: http://b/142666190

### DIFF
--- a/other-builds/ndkbuild/bitmap-plasma/app/Android.mk
+++ b/other-builds/ndkbuild/bitmap-plasma/app/Android.mk
@@ -1,4 +1,5 @@
-# Lopyright (C) The Android Open Source Project
+#
+# Copyright (C) 2016 The Android Open Source Project
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


DanAlbert:
(one sec, didn't notice that this was duplicated across more than one android.mk, will send an update)

ggfan:
PR already there https://github.com/android/ndk-samples/pull/657
when NDK-R21 is formally released, it will be merged

DanAlbert:
Updated.

> PR already there #657

This one adds comments explaining why and dedups everything.

> when NDK-R21 is formally released, it will be merged

Why not merge it now? It's compatible with older releases. Master is what Studio pulls from with "Import sample". I'm guessing no one looks at other branches.

ggfan:
The existing samples tends to support the formal releases.  this change is fine with NDK r16, we could merge. 

If you are changing the license line, might just change to this year, not 2016?

DanAlbert:
The samples need to support the betas. People use the betas.

AIUI the copyright should be the date of file creation, but I'm not an expert on that.

ggfan:
make sense, let's merge it.

PR #667, opened 2019-11-19
Fix file and class names in endless-tunnel readme

emrekultursay:
Encountered some nits in the doc while playing with the sample.


ggfan:
something wrong with CI, will check it out separately

PR #678, opened 2019-12-16
Add sample for using C++ dependencies from AARs.

DanAlbert:
A sample showing how to import native dependencies from Maven. This sample uses curl, OpenSSL, and JsonCpp to display a list of the most recent 10 reviews submitted to AOSP's code review system.

This is still somewhat WIP (it needs a lot of comments, and shouldn't be checked in until it's usable with a published AGP), but it's ready for some initial review.


DanAlbert:
@enh-google could you review the JNI/C++ here?

prefab/curl-ssl/app/src/main/cpp/java_interop.h
```diff
@@ -0,0 +1,80 @@
...
+
+    JniString cstr(env->GetStringUTFChars(from, nullptr),
+                   [=](const char* p) { env->ReleaseStringUTFChars(from, p); });
+
+    if (cstr == nullptr) {
+      LOG_ERROR("%s: GetStringUTFChars failed", __func__);
+      std::abort();
+    }
+
+    const jsize len = env->GetStringUTFLength(from);
```
enh-google: i think this is actually more expensive than just letting std::string do the strlen(), so not worth the extra line.

prefab/curl-ssl/app/src/main/cpp/java_interop.h
```diff
@@ -0,0 +1,80 @@
...
+    return env->NewStringUTF(from.c_str());
+  }
+};
+
+template <typename JavaSubType>
+struct Convert<jobjectArray, JavaSubType> {
+  template <typename SubType>
+  static jobjectArray from(JNIEnv* env, const std::vector<SubType>& from) {
+    jobjectArray array = env->NewObjectArray(
+        from.size(), env->FindClass(JavaClass<JavaSubType>::name), nullptr);
```
enh-google: (i assume you don't care about missing error checks, since you explicitly abort() later. i think they'd just make the code harder to read for little value.)

prefab/curl-ssl/app/src/main/cpp/app.cpp
```diff
@@ -0,0 +1,100 @@
...
+  };
+
+  const char mainClass[] = "com/example/curlssl/MainActivity";
+  jclass c = env->FindClass(mainClass);
+  if (c == nullptr) {
+    LOG_ERROR("Could not find %s", mainClass);
+    return JNI_ERR;
+  }
+
+  int rc = env->RegisterNatives(c, methods,
```
enh-google: why bother with RegisterNatives? just let the VM use dlsym() for less code?

prefab/curl-ssl/app/src/main/cpp/java_interop.h
enh-google: just inline "java/lang/String" here and remove the unneeded generality of `JavaClass`?

prefab/curl-ssl/app/src/main/cpp/macros.h
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#define DISALLOW_COPY_AND_ASSIGN(TypeName) \
```
enh-google: you only have one user. inline?

prefab/curl-ssl/app/src/main/cpp/http.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include <optional>
+#include <string>
+
+#include "curl/curl.h"
+
+using namespace std::string_literals;
+
+namespace {
+
+static std::string buffer;
```
enh-google: i'd be tempted to either get rid of the class and just have a C function that does everything, or at least fix this if you do intend this to be a useful bit of copy & paste?

prefab/curl-ssl/app/src/main/cpp/app.cpp
```diff
@@ -0,0 +1,100 @@
...
+  }
+  return titles;
+}
+
+}  // namespace
+
+JNIEXPORT jobjectArray JNICALL get_changes_jni(JNIEnv* env, jobject /* this */,
+                                               jstring cacert_java) {
+  if (cacert_java == nullptr) {
+    LOG_ERROR("cacert argument cannot be null");
```
enh-google: `env->FatalError()` for all of these logs/abort()s?

prefab/curl-ssl/app/build.gradle
```diff
@@ -0,0 +1,73 @@
...
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation"org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation 'com.android.support:appcompat-v7:29.0.0'
+    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
+    // NB: These are not published to Maven yet. You can build them yourself with
+    // https://cs.android.com/android/platform/superproject/+/master:tools/ndkports/.
+    implementation "com.android.ndk.thirdparty:curl:7.66.0"
+    implementation "com.android.ndk.thirdparty:jsoncpp:1.9.1"
+    implementation "com.android.ndk.thirdparty:openssl:1.1.1d"
```
ChristianRomberg: Hi,
is there any progress on this? Is there a way to help get these packages published?

DanAlbert: Right now we're waiting for all the process in our end to resolve. There isn't a version of the Gradle plugin capable of using these yet either. Hopefully by the end of the month we'll have both of those available.

prefab/curl-ssl/app/src/main/cpp/java_interop.h
DanAlbert: Not sure I understand. Not sure where the abort that would catch this is. I added the error check. The purpose of this header is to keep this kind of cruft out of the rest of the code, so I think it's a good place to add the check.

prefab/curl-ssl/app/build.gradle
DanAlbert: They're published now. The latest version of this PR has the correct package names and versions.

DanAlbert:
> i'd be tempted to optimize for either "absolute minimum, so you can see what's what" or "something you can copy & paste directly". and i'm assuming that the latter might want to be its own AAR anyway, so the sample should be the former?

I think that's what I have here. I could reduce the number of files or functions, but I'd just be copy pasting the code that's here into a different, larger function. I think it's much easier to read if it's factored out.

It would definitely be nice to get some of this (in particular the Java interop bits) out into its own AAR, but since that's a non-trivial effort it has to stay here until I can finish doing that.

prefab/curl-ssl/app/src/main/cpp/java_interop.h
```diff
@@ -0,0 +1,69 @@
...
+struct Convert<jobjectArray, jstring> {
+  static jobjectArray from(JNIEnv* env, const std::vector<std::string>& value) {
+    const char stringClassName[] = "java/lang/String";
+    jclass stringClass = env->FindClass(stringClassName);
+    if (stringClass == nullptr) {
+      logging::FatalError(env, "%s: FindClass(\"%s\") failed", __func__,
+                          stringClassName);
+    }
+
+    jobjectArray array = env->NewObjectArray(value.size(), stringClass, nullptr);
```
enh-google: NewObjectArray can fail.

prefab/curl-ssl/app/src/main/cpp/java_interop.h
```diff
@@ -0,0 +1,69 @@
...
+    for (size_t i = 0; i < value.size(); ++i) {
+      env->SetObjectArrayElement(array, i, env->NewStringUTF(value[i].c_str()));
```
enh-google: as can NewStringUTF.

(SetObjectArrayElement can too, but since that's a logic error, i think just letting the VM abort is fine.)

DanAlbert:
Doesn't build in CI because the buildbot doesn't have a new enough version of gradle for the 4.0 AGP. Backed out the CI change for now.

PR #679, opened 2019-12-18
Update display-p3 sample

ggfan:
with EGL_EXT_gl_colorspace_display_p3_passthrough

Test: Pixel 3 XL Android 10

### DIFF
--- a/display-p3/image-view/src/main/cpp/WideColorCtx.cpp
+++ b/display-p3/image-view/src/main/cpp/WideColorCtx.cpp
@@ -27,12 +27,27 @@ struct GL_WIDECOLOR_MODE_CFG {
 };
 
 static const APP_WIDECOLOR_MODE_CFG appWideColorCfg[] = {
+    { DISPLAY_COLORSPACE::P3_PASSTHROUGH, DISPLAY_FORMAT::R8G8B8A8_REV },
+    { DISPLAY_COLORSPACE::P3_PASSTHROUGH, DISPLAY_FORMAT::R10G10B10_A2_REV},
+    { DISPLAY_COLORSPACE::P3_PASSTHROUGH, DISPLAY_FORMAT::RGBA_FP16},
     { DISPLAY_COLORSPACE::P3, DISPLAY_FORMAT::R8G8B8A8_REV },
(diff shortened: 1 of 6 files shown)


ggfan:
@courtney-g may you take a look?  thank you!

display-p3/image-view/src/main/cpp/WideColorCtx.cpp
```diff
@@ -158,30 +172,49 @@ bool ImageViewEngine::CreateWideColorCtx(void) {
...
 
+  /* check for GL_EXT_gl_colorspace_display_p3_passthrough supportability:
+   * implemented from Android 10+.
+   * In this sample, when p3_passthrough_ext is enabled, the texture content needs to
+   * be display ready (no EOTF transfer function will apply ), hence app needs to tell texture
+   * is in LINEAR mode so sampler will NOT apply OETF transfer;
+   * on the other hand, if p3_passthrough_ext is unavailable, app would use p3_ext;
+   * in this case, all hardware transfer function applies ( OETF & EOTF ), so app
+   * informs gl engine that the textures are in SRGBA format.
+   */
```
olvaffe: The meanings of EOTF and OETF seem reserved here.

Image files store pixel values with OETF (optical-electro transfer function) applied.  It is like a compression.  Apps can control whether the sampler should apply EOTF to return decompressed pixel values or not.  Similarly, apps can control whether the color output unit should apply OETF to write compressed pixel values to the color buffers.  Mis-configuring GLES might result in double-compressed pixel values and the like.

Also, the system compositor expects compressed pixel values when the colorspace is P3 or P3_PASSTHROUGH.  Apps must make sure they write compressed pixel values to the color buffers.

courtney-g: Agree, use of EOTF and OETF are off. I always have a tough time keeping them straight as well.
OETF means that the image data has been transformed from optical (O)(linear space) to electronic (E) (gamma / compressed space). It needs to be in this form with an sRGB transfer function applied when sent to a typical sRGB display.
Otherwise looks okay.

ggfan:
thank you, applied the fix!

PR #681, opened 2020-01-29
Add a hwasan sample.

eugenis:
Shows build system changes to enable hwasan on an app, and a bunch of
buttons to trigger different types of crashes.


hello-hwasan/app/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
DanAlbert: Do we have a higher level `.gitignore` that's catching things like `app/build` and `app/.cxx`?

hello-hwasan/README.md
```diff
@@ -0,0 +1,57 @@
...
+1. Click *Tools/Android/Sync Project with Gradle Files*.
+1. Click *Run/Run 'app'*.
+
+Screenshots
+-----------
+![screenshot](screenshot.png)
+
+Support
+-------
+If you've found an error in these samples, please [file an issue](https://github.com/googlesamples/android-ndk/issues/new).
```
DanAlbert: This URL will redirect here, but we should probably try to use android/ndk-samples rather than googlesamples/android-ndk when we're adding new samples (although @ggfan is the owner here, so it's up to him)

hello-hwasan/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Patches are encouraged, and may be submitted by [forking this project](https://github.com/googlesamples/android-ndk/fork) and
```
DanAlbert: Ditto

hello-hwasan/app/build.gradle
```diff
@@ -0,0 +1,49 @@
+apply plugin: 'com.android.application'
+
+apply plugin: 'kotlin-android'
+
+apply plugin: 'kotlin-android-extensions'
+
+android {
+    compileSdkVersion 28
```
DanAlbert: 29?

hello-hwasan/app/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+    buildToolsVersion "29.0.2"
+    defaultConfig {
+        applicationId "com.example.hellohwasan"
+        minSdkVersion 28
```
DanAlbert: 29, right?

hello-hwasan/app/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+        targetSdkVersion 28
```
DanAlbert: 29

hello-hwasan/app/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+        versionCode 1
+        versionName "1.0"
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        externalNativeBuild {
+            cmake {
+                cppFlags ""
```
DanAlbert: omit

hello-hwasan/app/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+        }
+    }
+
+    externalNativeBuild {
+        cmake {
+            path "src/main/cpp/CMakeLists.txt"
+            version "3.10.2"
```
DanAlbert: Omit this. The gradle plugin already locks the version for us, so the only thing putting this here does is accidentally keep us on an old version of CMake when we update AGP and forget to update this. That's not something we want people cargo culting into their own builds. (All of the existing samples probably need to be cleaned up, but let's not spread the problem in new samples.)

hello-hwasan/app/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+}
+
+dependencies {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation"org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
```
DanAlbert: Space between `implementation` and the argument (weird that this compiles)

ggfan:
samples tends to show how to use APIs, hwasan is the tool, it is more suitable to have great documentation on developer.android.com?  things like delete a pointer twice prevention may not be best for the samples.

hello-hwasan/app/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+cmake_minimum_required(VERSION 3.4.1)
+add_library(native-lib SHARED native-lib.cpp)
+
+if(ANDROID_ABI STREQUAL "arm64-v8a")
```
DanAlbert: If we only support this on arm64, restrict the sample to arm64 only in the build.gradle using [ABI filters](https://developer.android.com/ndk/guides/abis#gc)

hello-hwasan/app/src/main/java/com/example/hwasantest/MainActivity.kt
```diff
@@ -0,0 +1,42 @@
...
+import android.os.Bundle
+import kotlinx.android.synthetic.main.activity_main.*
+
+class MainActivity : AppCompatActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_main)
+
+        button_uaf.setOnClickListener { _ ->
```
DanAlbert: The `_ ->` is redundant in all of these. Just

```kotlin
button.setOnClickListener {
    doBlah()
}
```

is sufficient

hello-hwasan/app/src/main/java/com/example/hwasantest/MainActivity.kt
```diff
@@ -0,0 +1,42 @@
...
+            doUseAfterFree();
```
DanAlbert: Omit semicolons in Kotlin

hello-hwasan/app/src/main/java/com/example/hwasantest/MainActivity.kt
```diff
@@ -0,0 +1,42 @@
+package com.example.hellohwasan
```
DanAlbert: Copyright header

hello-hwasan/app/src/main/cpp/native-lib.cpp
```diff
@@ -0,0 +1,64 @@
+#include <jni.h>
```
DanAlbert: Copyright header in source files (aiui they're optional in build files, but should be in source files)

DanAlbert:
> samples tend to show how to use APIs, hwasan is the tool, it is more suitable to have a great documentation on developer.android.com? things like deleting a pointer twice prevention may not be best for the samples.

It's probably useful to have samples for people to play with for tools too. If people are having trouble getting the tools working it's useful to be able to check out a known good configuration to see what you're doing wrong.

If you want to keep them separate, what do you think of separating samples into `apis` and `tools` subdirectories?

PR #682, opened 2020-02-01
Fix Clang Tidy for native-activity

XieEDeHeiShou:


XieEDeHeiShou:
Should I format main.cpp with the default cpp code format?

native-activity/app/src/main/cpp/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ cmake_minimum_required(VERSION 3.4.1)
 # build native_app_glue as a static lib
 set(${CMAKE_C_FLAGS}, "${CMAKE_C_FLAGS}")
 add_library(native_app_glue STATIC
-    ${ANDROID_NDK}/sources/android/native_app_glue/android_native_app_glue.c)
+        ${CMAKE_ANDROID_NDK}/sources/android/native_app_glue/android_native_app_glue.c)
```
ggfan: ANDROID_NDK is the from Android port of CMake, may you share your issues with it?

XieEDeHeiShou: There's nothing wrong with it.
[The CMake simple highlighter plugin](https://plugins.jetbrains.com/plugin/10089-cmake-simple-highlighter/) told me this variable may not be initialized, and the quick fix provides the `CMAKE_ANDROID_NDK` variable.

TIM截图20200205113228

TIM截图20200205113613

Ref: https://cmake.org/cmake/help/v3.7/manual/cmake-toolchains.7.html#cross-compiling-for-android

XieEDeHeiShou: Should I revert this change ?

ggfan: it might be good to keep ANDROID_NDK as it is now ( kind of consistent with Android.MK's value ). thank you

ggfan:
thank you for your pull request! merged now!

PR #686, opened 2020-02-14
Add a simpler Prefab sample.

DanAlbert:
This sample only depends on jsoncpp, and trivially extracts a value
from a JSON string passed from Java.


ggfan:
Intending to retire this PR with a real NDK example: https://github.com/android/ndk-samples/pull/714, please review there.  Thanks for the PR

ggfan:
I forked and will update it, then create PR.

ggfan:
This PR is aliased to https://github.com/android/ndk-samples/pull/752 now, thanks.

PR #704, opened 2020-04-02
Update Hello JNI dependencies and fix lint issues.

DanAlbert:
Fixes http://b/153108278

### DIFF
--- a/hello-jni/app/build.gradle
+++ b/hello-jni/app/build.gradle
@@ -5,7 +5,7 @@ android {
     defaultConfig {
         applicationId 'com.example.hellojni'
         minSdkVersion 23
-        targetSdkVersion 28
+        targetSdkVersion 29
         versionCode 1
         versionName "1.0"
     }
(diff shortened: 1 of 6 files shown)


hello-jni/app/build.gradle
```diff
@@ -18,47 +18,13 @@ android {
     }
     externalNativeBuild {
         cmake {
-            version '3.10.2'
             path "src/main/cpp/CMakeLists.txt"
         }
     }
-
-    flavorDimensions 'cpuArch'
```
ggfan, line 26: this one is to demo productFlavors Options, probably keep?
32 bit vs. 64 bit, Android TV is still on 32 bit, play store is requesting new apps to be 64 bit only. 

we want samples to demonstrate various settings to build applications, this might keep it.

DanAlbert: This is demoing how to build unusable APKs though, since they don't conform to the play store 64-bit requirements.

Either way, what's a productFlavors demo doing in hello world?

ggfan: I thought this would allow just build on flavor of the product inside IDE and with Gradle on command line, and would only include one flavor ( in this case, CPU type ) in the APK. ( otherwise it would build everything in IDE ).   The samples here also used by other downstream project like Android TV ( not relevant, they were just complaining about why we are not release 32 bit version of other android build project yesterday -- we were only releasing 64 bit versions)

ggfan: yeah, that is true: play store would require 64 bit lib inside the archive/apk; this 32 bit cpu only flavor would only have 32 bit lit in the apk.

ggfan:
updated the code with a new PR:  https://github.com/android/ndk-samples/pull/706
retiring this one: app build but could run with this one, refer to the new PR for details.

PR #707, opened 2020-04-14
fix memcpy error when stride > width

chenjim:
memcpy size is width  so  src += width


ggfan:
closing this one for now, please let me know if you think otherwise!  thank you for the attention to the samples!

PR #708, opened 2020-04-15
Prefab/hello-libs:  for prefab debugging, do not merge

alexcohn:
use prefab for hello-libs


ggfan:
thank you for the Pull Request!  I will have a look later: given that there is one sample already, and the other one from danalbert@ is inside Pull Request, this one probably is not an urgent one.  The uniqueness is about the generating the lib part, I think. Let me know you think  otherwise!

alexcohn:
> The uniqueness is about the generating the lib part

That's true; also, the important message is to use **Prefab** for consuming external libs, whenever possible.

But actually, this sample is intended to triage some features of Prefab (e.g. reusing library bundle for prefab).

ggfan:
this is to replicate issue:  https://b.corp.google.com/issues/120166563

ggfan:
https://issuetracker.google.com/issues/120166563 is under implementation, let's follow it up there.
thank you for creating the test case.

PR #720, opened 2020-04-28
Clean up -Wunused-variable and -Wimplicit-fallthrough warnings.

bashi-bazouk:


display-p3/image-view/src/main/cpp/simple_png.cpp
```diff
@@ -130,12 +130,17 @@ PNGHeader::PNGHeader(std::string& name, uint8_t *buf, uint64_t len) :
       default:
         LOGV("====Unprocessed CHUNK %c%c%c%c",
              type.bytes[3], type.bytes[2], type.bytes[1], type.bytes[0]);
-        // on purpose: pass through
+        [[fallthrough]];
```
ggfan, line 133: this requires C++ 17,  could you change 11 to 17 in [CMakeLists.txt](https://github.com/android/ndk-samples/blob/master/teapots/image-decoder/src/main/cpp/CMakeLists.txt#L40)

webp/view/src/main/cpp/webp_view.cpp
```diff
@@ -189,6 +189,7 @@ bool Engine::PrepareDrawing(void) {
             break;
         case WINDOW_FORMAT_RGBA_8888:
             descriptor.format_ = SurfaceFormat::SURFACE_FORMAT_RGBA_8888;
+            [[fallthrough]];
```
ggfan, line 192: add a `break;` instead: should be supportable; otherwise
change the build to request for 17 [here]()https://github.com/android/ndk-samples/blob/master/webp/view/src/main/cpp/CMakeLists.txt#L47?

Issue #722, opened 2020-04-29
this whole project need to be updated


Niek-Okido:
hi,

the project teapot and some more are really out dated the sdk version is so old that it is not able to open project in android studio (on the normal way)

hope you get this updated,

kind regards

ggfan:
do you mean things like:
```
    compileSdkVersion 'android-R'
    buildToolsVersion '30.0.0 rc2'
```
those are in canary, do you have canary update channel enabled?

Niek-Okido:
yeah i have (so far as i know), 

`ndkVersion '21.1.6210238'`
this version is not anymore included in android-ndk
the thing is when you open the project he gives me so much errors.

why you not just update it to the latest ndk version, than at least a lot more people are able to use it.

greetings,

Niek-Okido:
i did check it, i use the beta version so need to update, thank you

ggfan:
just to confirm with you: after you updated with the canary channel, you could compile, correct?

currently the NDK for Android R developer preview could only be installed via sdkmanager ( UI or commandline ), unavailable on public download site, check [wiki page](https://github.com/android/ndk-samples/wiki/Configure-NDK-Path) for some background info when necessary.

Looks like that Android Preview and Beta features need to be put into a new beta-sample branch to avoid confusions: this came up a few times this year. Will look to fix it at the future Android R release milestone.

Niek-Okido:
i will try it, will update the result 

thanks,

Niek-Okido:
i did try what you said, but still some issue :
`> No version of NDK matched the requested version 21.1.6210238. Versions available locally: 21.0.6113669, 21.1.6352462, 21.1.6363665`
and i try to get this version but no success
any advice.

and can you not make this project(s) available for the stable version (very irritating)

kind regards

ggfan:
updated the ndk to the latest stable version, should fixed the issue you see.

The warning you see though might just a false warning: it actually is downloading that version for you, just the false warning.

Niek-Okido:
now i get as well some errors like `Task 'wrapper' not found in project ':choreographer-30fps'.` but yeah i leave this now because not working .

i can bet a lot of more people just skip the ones that don't work

kind regards,

ggfan:
should be fixed now: the beta features are inside "develop" branch

PR #723, opened 2020-05-01
Fix NDK version for textured teapot.

ggfan:


ggfan:
this should fix https://github.com/android/ndk-samples/issues/722

PR #724, opened 2020-05-12
Remove obsolete jniLibs configuration.

DanAlbert:
Configuring jniLibs for explicit dependencies is no longer necessary
as of 4.0. In fact, this will result in a build failure since it is
being copied twice:

    * What went wrong:
    Execution failed for task ':app:mergeDebugNativeLibs'.
    > A failure occurred while executing com.android.build.gradle.internal.tasks.Workers$ActionFacade
       > More than one file was found with OS independent path 'lib/x86/libgperf.so'

Also see http://b/154909008

### DIFF
--- a/hello-libs/README.md
+++ b/hello-libs/README.md
@@ -23,9 +23,6 @@ When importing libraries into your app, include the following in your app's `CMa
 *    configure each library binary location(using `set_target_properties`)
 *    configure each library headers location (using `target_include_directories`)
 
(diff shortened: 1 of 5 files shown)


ggfan:
Oh, it is 4.0+ feature, 3.6.3 still needs the correct manual packing.

ggfan:
updated this PR to https://github.com/android/ndk-samples/pull/735, retiring this one.

PR #729, opened 2020-05-31
Update all projects to Stable AGP 4.0.0

nisrulz:
Note: Minimum gradle wrapper required is now 6.1.1, so that is also updated. 
Take a look at commits for each change.


ggfan:
could you rebase the pull request? the build error for hello-libs should be fixed in main branch.  Thank you!

PR #734, opened 2020-06-01
Delete build.gradle

jackevill:


ggfan:
there might be some misunderstanding: this file it necessary to build this example.

PR #749, opened 2020-09-14
Improve ndk-samples / prefab

Consti10:
This pull request contains 2 changes:
a) The dummy method of prefab-publishing now returns a dummy std::string. 
This string can be used in prefab-dependency to create a meaningfull example
b) New project prefab-dependency:
Add an example project that depends on the prefab-publishing library. Its history is as follow:
1) Create a native c++ application via the android studio wizzard
2) Add the prefab-publishing module as a dependency (import .aar)
3) Display the dummy string obtained via my_api from prefab-publishing.


Consti10:
@googlebot I fixed it.

DanAlbert (changes requested):
First, thanks a ton for doing this. I know this is a really difficult workflow in AGP right now.

This uses a checked in prebuilt mylibrary-debug.aar, so this isn't so much a way of showing how to use an inter-module dependency as it is an example showing how to avoid maven. Until https://issuetracker.google.com/120166563 is resolved there aren't really other options, but I don't think this sample really makes that clear unless you already know why it's needed. What's being done here is probably much more clear in doc form.

If we do want a sample that shows what you can't do, I think this could do that with some changes:

* Remove `TelemetryCore`. It's unused.
* Remove the prebuilt AARs
* Merge the mylibrary source into `prefab-dependency/mylibrary`. Use a gradle property to control whether the library should be built or if the prebuilt should be used. Use the prebuilt by default, but make it so `./gradlew -PbuildMyLibrary :mylibrary:assemble` (or whatever the task name is) works
* Add a readme to explain the purpose of the sample, the workflow required, and a link to the bug describing the future improvement
* For the prebuilt flow, add a check to `mylibrary/build.gradle` to check if `mylibrary-debug.aar` exists, and raises an error pointing the user at the readme if it does not

Also probably (@ggfan correct me if I'm wrong) replace the package names used to be something generic.

I'm also completely okay with none of that work getting done and instead converting this to a doc. It'd be a lot easier, and we can link it in the other samples so its discoverable here.

ggfan:
Give me some time to look into this one and figure out a path. thank you both for looking into this one!

Consti10:
Avoiding maven was not really my intention, but since jitpack does not work with latest gradle yet I decided to skip that part.
I know it is also a little bit pointless to import a generated .aar when the imported project is in the same github project :) - therefore you could say prefab-dependency is more of a test using prefab-publishing than an example.

Consti10:
For an example it probably makes more sense to add Jitpack or maven to the prefab-publishing project and let the prefab-dependency project fetch the .aar online.

DanAlbert:
What would that demonstrate that the two existing samples don't?

Consti10:
curl and ssl are not published via prefab-publishing, right ? I think there are a lot of developers that are going to use the android prefab-publishing gradle plugin for their project rather than ndk-ports

DanAlbert:
I agree that's what most users will use, but the way the package is generated doesn't affect how it is used, so I'm confused as to what this shows the user that wasn't already shown.

Consti10:
You are right, it doesn't show anything 'new' to the user. However, I would have found it helpfull if there was a project I can just clone and then do modifications to both the published library and the project that depends on the published library. Since curl / ssl is built at another place the curl-ssl example cannot be used in that way.

ggfan:
first, apologize for the delay: taking some time to sort this out. Now with "some" prefab background :-), I think that the PR has value of:

-  how to configure a local AAR into app's dependency ( bypassing downloading from maven )
-  Native AAR is one kind of AAR, once it is within the dependency, prefab could access the native modules inside the AAR

It is a valid use case, and probably very useful for AAR generating phase, for example, for test purpose before uploading to maven.  it is handy, thank you for bring it up!

By looking at the [existing AAR documentation](https://developer.android.com/studio/projects/android-library),  your example way could be added into it: it is generic for all AARs.

Once the dependency is added, the app's side configuration to use native lib inside AAR is the same as the "download way" shown inside the existing sample called [curl-ssl](https://github.com/android/ndk-samples/tree/main/prefab/curl-ssl). Probably the cur-ssl sample could be changed to something more prefab advocating like  what you are using, "prefab-dependency".

I would expand AAR generic documentation to include this sample's way to add AAR dependency; then change the same name from curl-ssl to prefab-dependency, rather than adding a new sample ( it does have a lot of overlap with the existing one ). 

thank you again for looking into prefab and improving sample quality and usage!

ggfan:
https://github.com/android/ndk-samples/pull/752 created to update README.md; formal documentations are updated  too:

- [Add AAR as dependency](https://developer.android.com/studio/projects/android-library#AddDependency)
- [C/C++ dependencies](https://developer.android.com/studio/build/native-dependencies?buildsystem=cmake&agpversion=4.0)

Now the 2 necessary steps are fully covered:

1. Add AAR remotely or locally to the project ( this is normal AAR thing )
2. Add native dependencies to app's native ( C/C++ ) code ( this is the value that the native AAR brings )

With that, I would close this PR and focus on the PR #752 .  Thank you again for your Pull Request: very helpful to improve samples and documentation!

Consti10:
Looks really good ! I hope more people start using prefab such that the number of native libraries for android on maven / jitpack increases.

ggfan:
Hey @Consti10 , just FYI:   the "prefab-dependency" went into the sample repo, your signature!
Cheers!

PR #777, opened 2021-02-18
Adding animated-image decoder sample under webp/ for Android DP1

ggfan:


webp/image-decoder/src/main/cpp/image_decoder.cpp
```diff
@@ -0,0 +1,240 @@
...
+        return false;
+
+    int status;
+    if(!curImgFile) {
+        curImgFile = fileNames.front();
+        fileNames.pop();
+        // Open the asset with the give name from the APK's assets folder.
+        imgAsset =
+                AAssetManager_open(assetMgr, curImgFile, AASSET_MODE_BUFFER);
+        ASSERT(imgAsset, "%s does not exist in %s", curImgFile,
```
philburk, line 93: Using ASSERT in an app is a problem because the app will crash and not recover from common errors like missing files.

ggfan: thanks for the comment: the sample at beta stage is to teach developers, if API failed, assert and stop right there to save debugging time and saving developer's time, the reported bug will be clean (rather than "the sample failed").

PR #791, opened 2021-04-16
Delete LICENSE

mrworaphat:


ggfan:
this file is legal requirement for repo level purpose(the copyright header in each file is for that file purpose), could not be removed.

PR #805, opened 2021-07-11
测试C调用java

leetiger:


ggfan:
Thank you for your Pull Request! Your PR seems to be an experiment to see how to callback a static member function, not directly add new usage to the sample itself though, like to pass.

PR #821, opened 2021-12-31
Large Screen / Chrome OS optimizations for audio-echo demo

MattMonasch:
Added landscape layout
Changed screenOrientation manifest tag to allow for different orientations
Added configChange manifest tag to preserve app state between window changes like orientation and form factor.
Added controller support.


MattMonasch:
If it's not already common practice, I'd like to squash this PR into a single commit.

PR #823, opened 2022-01-06
Add a sample showing how to use PGO.

DanAlbert:
This shows how to use profile-guided-optimization (PGO) in an NDK app.
This is a two step process, where the developer runs a sample workload
in profiling mode, fetches the profile data from the device, then
rebuilds the project using that profile. Profiles are somewhat stable
so can be reused as the code changes, but should be updated
periodically.

This is probably worth some explicit support in Android Studio, but
until that happens this shows how to do it by hand. Docs to accompany
this sample are also being worked on (in critique).


jfgoog (approved):
Cool!

PR #849, opened 2022-07-01
Updating hello-jnicallback to support large screen devices

MattMonasch:
Timer does not restart on configuration changes and utilizes a ViewModel to preserve state. Timer functionality refactored into this ViewModel.
Button added to start and stop.
Changed StopTimer to stopTimer to match naming convention.


ggfan (approved):

DanAlbert:
Sorry for the delay, there wasn't anyone to look after the samples for a while.

What does this actually fix? What's a "large screen device"? I don't know how to verify that this does what it says.

DanAlbert:
I can see that one thing it does fix is that the timer doesn't restart when the screen rotates, but I'm still not sure what this has to do with "large screen devices", whatever those are.

PR #866, opened 2022-11-17
Upate gradle files to reflect current app templates.

jfgoog:
settings.gradle:
* Add pluginManagement and dependencyResolutionManagement stanzas.

build.gradle:
* Remove declaration of repositories and dependencies, since this goes in settings.gradle
* Use Kotlin version 1.7.10
* Use AGP version 7.3.0

app/build.gradle:
* Use `plugins { id 'foo' }` instead `apply plugin: 'foo'`
* Compile all samples with the latest version of the SDK (33) and NDK (25.1.8937393).
* Remove kotlin-android-extensions

Furthermore:
* In CI build script, make detection of debug mode work on MacOS.
* Uncomment webp in CI build script.
* Add missing projects to CI build script (display-p3, hello-oboe, native-midi, prefab/prefab-dependency)
* Update gradle version to 7.4
* Update hello-oboe and prefab-dependency to use view binding.
* Update native-midi to eliminate compiler warning.

### DIFF
--- a/native-midi/app/src/main/cpp/AppMidiManager.cpp
+++ b/native-midi/app/src/main/cpp/AppMidiManager.cpp
@@ -148,13 +148,12 @@ extern "C" {
 void Java_com_example_nativemidi_AppMidiManager_startReadingMidi(
         JNIEnv* env, jobject, jobject midiDeviceObj, jint portNumber) {
 
-    media_status_t status;
-    status = AMidiDevice_fromJava(env, midiDeviceObj, &sNativeReceiveDevice);
+    AMidiDevice_fromJava(env, midiDeviceObj, &sNativeReceiveDevice);
     // int32_t deviceType = AMidiDevice_getType(sNativeReceiveDevice);
     // ssize_t numPorts = AMidiDevice_getNumOutputPorts(sNativeReceiveDevice);
 
     AMidiOutputPort *outputPort;
-    status = AMidiOutputPort_open(sNativeReceiveDevice, portNumber, &outputPort);
+    AMidiOutputPort_open(sNativeReceiveDevice, portNumber, &outputPort);
 
     // sMidiOutputPort.store(outputPort);
     sMidiOutputPort = outputPort;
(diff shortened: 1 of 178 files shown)


native-midi/app/src/main/cpp/AppMidiManager.cpp
```diff
@@ -148,13 +148,12 @@ extern "C" {
 void Java_com_example_nativemidi_AppMidiManager_startReadingMidi(
         JNIEnv* env, jobject, jobject midiDeviceObj, jint portNumber) {
 
-    media_status_t status;
```
DanAlbert, line 151: Shouldn't errors at least be logged?

DanAlbert (changes requested):

native-midi/app/src/main/cpp/AppMidiManager.cpp
jfgoog: This is purely to silence a `-Werror` from a variable that is set but never used. There should be no change in behavior.

DanAlbert: I know, but the change just makes the symptom go away, not the problem. Is there some reason we shouldn't be handling these error cases? If so, that's worthy of a comment.

jfgoog: Well, I didn't look too deeply, but there didn't seem to be an obvious way to do something better with the error.

Now, as I look more closely, this is already not the only instance of swallowing statuses. For example, I just spotted this gem on line 177:

`/*media_status_t status =*/ AMidiDevice_release(sNativeReceiveDevice);`

I mean, I can add a TODO if you want, but actually improving the functionality of the samples is not the point of this change, and I'm not ready to go down that rabbit hole just yet.

DanAlbert: Filing a bug would work. I just know that if we don't do anything, we will forget about this.

jfgoog: https://github.com/android/ndk-samples/issues/870

DanAlbert (approved):

PR #868, opened 2022-11-24
Adding global .clang-format from Khronos

GOOG-sergiu:
I am currently in the process of writing a new sample (PR going up soon) and realised there's no unified code style for this repository?

@ggfan suggested to use this .clang-format from Khronos for the whole repository.
https://github.com/KhronosGroup/Vulkan-Samples/blob/master/.clang-format

I sort of wouldn't want to format all the samples and get my name plastered over all of them, but I think going forward there's a need to have a unified style and we can enforce this one for all new samples? Thoughts? 

I can be persuaded to change all of them, but I think the official maintainers for the samples are better placed to do the changes for their respective samples, if we want to update all of them?

I feel this will be a contentious point so I'll leave this up for discussion :)

Saying that, I am not sure if this doesn't clash with:
/prefab/.clang-format ? 

Is that the style used for all the other samples? In which case we can simply use that.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,106 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
+AlignConsecutiveDeclarations: true
+AlignEscapedNewlines: Left
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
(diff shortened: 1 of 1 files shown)


DanAlbert:
There was a clang-format CI step for a while, but it proved more or less impossible to maintain. clang-format's behavior does change from release to release (bug fixes and improvements to clang-format both result in a style change), and without a way to synchronize clang-format versions across developers, it was making PRs very difficult to merge.

Agreed that we ought to have one even if it's not auto-enforced though. Also agreed that it should just be the Google style. https://github.com/android/ndk-samples/pull/1001 promotes prefab's to the top level. Closing this in favor of that.

PR #874, opened 2022-11-30
Update .ci_tools/set_env.sh to support 'CompileSdkVersion` too.

ggfan:
The existing samples use `compileSdk` in build.gradle; looks like `compileSdkVersion` is also supported, need to handle it in set_env.sh. Also try to use the new location (if it is exists) for sdkmanager command line tool, PTAL.

### DIFF
--- a/.ci_tools/setup_env.sh
+++ b/.ci_tools/setup_env.sh
@@ -4,6 +4,7 @@
 #  - pwd must be inside the repo's homed directory (android-ndk)
 #  - upon completion, we are still in the same directory ( no change )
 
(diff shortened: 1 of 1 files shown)


GOOG-sergiu (approved):
LGTM 👍 Thanks!

jfgoog:
https://github.com/jfgoog/ndk-samples/commit/10439ab36371e007532a73b5a2dcb593cc04e8fb should have changed compileSdkVersion to compileSdk everywhere. I believe `compileSdk` is now the preferred syntax, and we don't need to support compileSdkVersion.

PR #908, opened 2022-12-06
Clang format

jfgoog:


matthias-blume (commented):
I assume the large batch of diffs is now because clang-format was applied?

jfgoog:
> I assume the large batch of diffs is now because clang-format was applied?

Correct. This is the result of running clang-format on everything, and is 100% automated.

DanAlbert (approved):
image

matthias-blume (approved):

PR #915, opened 2022-12-08
Unit test

jfgoog:


unit-test/app/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+        $<TARGET_OBJECTS:adder>
+        native-lib.cpp)
+
+# Searches for a specified prebuilt library and stores the path as a
+# variable. Because CMake includes system libraries in the search path by
+# default, you only need to specify the name of the public NDK library
+# you want to add. CMake verifies that the library exists before
+# completing its build.
+
+find_library( # Sets the name of the path variable.
```
DanAlbert, line 38: I don't know why all the existing samples insist on this pattern. `target_link_library(blah PRIVATE log)` works just fine.

unit-test/app/src/main/cpp/adder.h
```diff
@@ -0,0 +1,6 @@
+#ifndef UNIT_TEST_ADDER_H
```
DanAlbert, line 1: `#pragma once`

DanAlbert (approved):

unit-test/app/src/main/cpp/adder.h
jfgoog: Done.

unit-test/app/src/main/cpp/CMakeLists.txt
jfgoog: This is from the Android Studio native app template. I'm going to leave it as-is.

PR #918, opened 2022-12-08
Speed up build.

jfgoog:


jfgoog:
Nope. Looks like we don't have nested virtualization. :(

jfgoog:
@DanAlbert this is basically FYI because it doesn't actually speed things up.

Honestly a little disappointed with the state of CI for Android apps.

Why doesn't Android dev rel maintain a GitHub action to install the SDK and NDK? The best one seems to be https://github.com/maxim-lobanov/setup-android-tools, which hasn't been updated in a couple of years.

Why is there no official framework for screenshot testing? https://facebook.github.io/screenshot-tests-for-android/ looks spiffy, but it doesn't work with the Gradle plugin DSL.

PR #919, opened 2022-12-12
Exceptions

jfgoog:


exceptions/app/src/main/cpp/native-lib.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include "exception_helper.h"
+
+extern "C" JNIEXPORT void JNICALL
+Java_com_example_exceptions_MainActivity_throwsException(JNIEnv* env,
+                                                         jobject /* this */) {
+  try {
+    throw std::runtime_error("A C++ runtime_error");
+  } catch (std::runtime_error e) {
+    jniThrowRuntimeException(env, e.what());
+  }
```
DanAlbert, line 14: It might be a better illustration here to move the throw to a function named something like `might_throw()`, and change the catch to just `std::exception`, since the goal is to prevent any exceptions from propagating across the JNI boundary (I suppose to be safe you should have the `std::exception` catch and also a `...` catch in case someone throws an `int` or whatever). A comment explaining that would also be good.

WDYT?

exceptions/app/src/main/java/com/example/exceptions/MainActivity.kt
```diff
@@ -0,0 +1,36 @@
...
+            throwsException();
+            binding.sampleText.text = "No exception thrown";
+        } catch (e: java.lang.RuntimeException) {
+            binding.sampleText.text = "RuntimeException caught. Message: \"" + e.message + "\"";
+        }
+
+    }
+
+    // Kotlin's Unit type corresponds to void in C/C++.
+    external fun throwsException(): Unit
```
DanAlbert, line 28:
```suggestion
    external fun throwsException()
```

jfgoog: Done

exceptions/app/src/main/cpp/native-lib.cpp
jfgoog: PTAL

exceptions/app/src/main/cpp/native-lib.cpp
```diff
@@ -4,12 +4,22 @@
...
 Java_com_example_exceptions_MainActivity_throwsException(JNIEnv* env,
                                                          jobject /* this */) {
   try {
-    throw std::runtime_error("A C++ runtime_error");
+    might_throw();
   } catch (std::runtime_error e) {
+    // Re-throw with the error message.
+    jniThrowRuntimeException(env, e.what());
+  } catch (std::exception e) {
+    // A more generic exception. We lose the specific message here.
```
DanAlbert, line 18: We don't actually lose anything, do we? The `std::runtime_error` catch block can be removed without any behavior change, I think.

DanAlbert (approved):

exceptions/app/src/main/cpp/native-lib.cpp
jfgoog: Yeah, you're right.

I just needed to catch by reference because, you know, inheritance is a thing.

Like I said, I don't know how to use exceptions in C++. :)

PR #920, opened 2022-12-12
Adding Google packaging.yaml to the hello-vulkan sample.

ggfan:


hello-vulkan/.google/packaging.yaml
```diff
@@ -0,0 +1,7 @@
+status:       PUBLISHED
+technologies: [Android, NDK, Vulkan, GameActivity]
```
jdkoren, line 2: I'm not sure `GameActivity` falls under technologies, it might be an API instead.

ggfan: removed, thank you for reviewing!

jdkoren (approved):

PR #921, opened 2022-12-13
Add mdformat pre-commit, and format all markdown.

jfgoog:


jfgoog:
I looked at the rich diff for all the files. Most have absolutely no changes (somewhat surprisingly). The 2 that do are minor and intended (there were bugs in the original markdown).

PR #925, opened 2023-01-12
Update samples in Android Studio.

jfgoog:
Remove the following from Android Studio, since they use deprecated APIs: audio-echo, native-audio.
Add the following samples to Android Studio: camera, exceptions, native-audio, native-codec, native-media, unit-test.

### DIFF
--- a/camera/.google/packaging.yaml
+++ b/camera/.google/packaging.yaml
@@ -3,5 +3,5 @@ technologies: [Android, NDK]
 categories:   [NDK]
 languages:    [C++, Java]
 solutions:    [Mobile]
-github:       googlesamples/android-ndk
+github:       android/ndk-samples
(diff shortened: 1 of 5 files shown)


DanAlbert (approved):
> they use deprecated APIs: audio-echo, native-audio.

Do we have samples that show how to use the replacement? Does the replacement sample work on Lollipop? If so, LGTM (approving because I think the answer is "yes, because oboe"). Otherwise we should probably keep them until _we_ have found the time to migrate.

jfgoog:
Yes, I think you are correct that the replacement is Oboe. But I'm just quoting the email you forwarded me on sample deprecation.

This is a "soft" delete -- they should disappear from the list in Android Studio, but they are still present in the repo.

PR #926, opened 2023-01-18
Sanitizer example

jfgoog:
Sanitizer example with build variants for asan and hwasan.


jfgoog:
Still TODO:
* README.md
* .google/packaging.yaml

sanitizers/app/build.gradle
```diff
@@ -0,0 +1,76 @@
...
+            version '3.18.1'
+        }
+    }
+    buildFeatures {
+        viewBinding true
+    }
+}
+androidComponents {
+    onVariants(selector().withBuildType("asan"), {
+        packaging.dex.useLegacyPackaging.set(true)
```
DanAlbert, line 63: Is there a bug that needs to be filed?

sanitizers/app/build.gradle
```diff
@@ -0,0 +1,76 @@
...
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+    }
+
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+        }
+        hwasan {
```
DanAlbert, line 26: Is it possible to use that build.gradle defined virtual device thing to automatically run this on a compatible system image?

sanitizers/app/src/asan/resources/lib/arm64-v8a/wrap.sh
```diff
@@ -0,0 +1,31 @@
+#!/system/bin/sh
+# Copied from https://android.googlesource.com/platform/ndk/+/refs/heads/master/wrap.sh/asan.sh
```
DanAlbert, line 2: Bonus points if you work out the gradle-fu to copy this from the NDK so it doesn't get stale (every so often we need to extend the jdwp workarounds to account for an ART change)

sanitizers/app/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+        SHARED
+
+        # Provides a relative path to your source file(s).
+        native-lib.cpp)
+
+if(SANITIZE)
+if(${SANITIZE} STREQUAL "hwasan")
+    message("Using hwasan")
+    target_compile_options(sanitizers PUBLIC -fsanitize=hwaddress -fno-omit-frame-pointer)
+    set_target_properties(sanitizers PROPERTIES LINK_FLAGS -fsanitize=hwaddress)
```
DanAlbert, line 30: You should be able to use https://cmake.org/cmake/help/latest/command/target_link_options.html these days. We have 3.22 available.

DanAlbert (approved):

sanitizers/app/src/main/cpp/CMakeLists.txt
jfgoog: Done.

sanitizers/app/src/asan/resources/lib/arm64-v8a/wrap.sh
jfgoog: Done.

sanitizers/app/build.gradle
jfgoog: Maybe. x86_64 and aarch64 both support hwasan. For now, I think I'm just going to point to the hwasan docs in developer.android.com.

PR #943, opened 2023-06-08
Remove redundant dex packaging option

emrekultursay:
The correct option is jniLibs packaging option, but that's not needed because it's already the default for debug builds.

### DIFF
--- a/sanitizers/app/build.gradle
+++ b/sanitizers/app/build.gradle
@@ -66,11 +66,6 @@ android {
         viewBinding true
     }
 }
-androidComponents {
-    onVariants(selector().withBuildType("asan"), {
-        packaging.dex.useLegacyPackaging.set(true)
-    })
-}
 
 dependencies {
 


emrekultursay:
@jfgoog I believe the `packaging.dex.useLegacyPackaging` option only applies to the dex files.

jfgoog:
I'm pretty sure this is actually required, because of the wrapper script that asan uses to launch.

Have you tested this change manually, on a physical device?

jfgoog:
See docs here: https://developer.android.com/ndk/guides/asan#running

emrekultursay:
The `jniLibs.packagingOptions.uselegacypackaging` and the `packaging.dex.useLegacyPackaging` option are two different options. 
https://developer.android.com/build/releases/past-releases/agp-4-2-0-release-notes#dex-files-uncompressed-in-apks-when-minsdk-=-28-or-higher

The jniLibs flavor is enabled by default when using debug builds (and that is why the sample was working fine for debug builds). Here's a screenshot of running ASan on a real device, with this section removed, and hitting a breakpoint just before it crashes due to address sanitizer.  https://screenshot.googleplex.com/3UyFdaNPadTftKP

jfgoog (approved):

Issue #929, opened 2023-01-29
native-activity crashes on startup


henke96:
The native-activity sample crashes on startup for me:
```
01-29 13:21:53.692 19506 19506 D AndroidRuntime: Shutting down VM
01-29 13:21:53.693 19506 19506 E AndroidRuntime: FATAL EXCEPTION: main
01-29 13:21:53.693 19506 19506 E AndroidRuntime: Process: com.example.native_activity, PID: 19506
01-29 13:21:53.693 19506 19506 E AndroidRuntime: java.lang.RuntimeException: Unable to get provider androidx.startup.InitializationProvider: java.lang.ClassNotFoundException: Didn't find class "androidx.startup.InitializationProvider" on path: DexPathList[[],nativeLibraryDirectories=[/data/app/~~4YAeCpMLzSkqGKIKZ4d1ng==/com.example.native_activity-dtqDfrF_hud_ev66GpcWfw==/lib/arm64, /data/app/~~4YAeCpMLzSkqGKIKZ4d1ng==/com.example.native_activity-dtqDfrF_hud_ev66GpcWfw==/base.apk!/lib/arm64-v8a, /system/lib64, /system_ext/lib64]]
01-29 13:21:53.693 19506 19506 E AndroidRuntime:        at android.app.ActivityThread.installProvider(ActivityThread.java:7478)
01-29 13:21:53.693 19506 19506 E AndroidRuntime:        at android.app.ActivityThread.installContentProviders(ActivityThread.java:6990)
01-29 13:21:53.693 19506 19506 E AndroidRuntime:        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6739)
01-29 13:21:53.693 19506 19506 E AndroidRuntime:        at android.app.ActivityThread.access$1500(ActivityThread.java:252)
01-29 13:21:53.693 19506 19506 E AndroidRuntime:        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2058)
01-29 13:21:53.693 19506 19506 E AndroidRuntime:        at android.os.Handler.dispatchMessage(Handler.java:106)
01-29 13:21:53.693 19506 19506 E AndroidRuntime:        at android.os.Looper.loopOnce(Looper.java:346)
01-29 13:21:53.693 19506 19506 E AndroidRuntime:        at android.os.Looper.loop(Looper.java:475)
[...]

PR #948, opened 2023-06-21
A workaround for issue#929

hybcloud:
according to issue https://github.com/android/ndk-samples/issues/929, native app cannot start successfully under such development tools and environment:

```
Android Studio Flamingo | 2022.2.1 Patch 2
Build #AI-222.4459.24.2221.10121639, built on May 12, 2023
Runtime version: 17.0.6+0-b2043.56-9586694 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Windows 11 10.0
GC: G1 Young Generation, G1 Old Generation
Memory: 1280M
Cores: 16
Registry:
    external.system.auto.import.disabled=true
    ide.text.editor.with.preview.show.floating.toolbar=false
    gradle.version.catalogs.dynamic.support=true
SDK: 33
NDK: 25.1.8937393
device: Samsung S970+
```
after this fix, the application can start and seems to perform correctly, but it might need someone experienced to check it twice so that it does not harm in respect of any other aspects.


hybcloud:
The problem is not solved 6 months after it was reported and given a "solution". I wish the issue could be taken into consideration asap.

DanAlbert:
Sorry, there hasn't actually been anyone able to look at this repo in quite some time. I'm doing what I can to volunteer some time for clean up.

A few other PRs (and at least one bug, I'm still sifting through and am sure I'll find more) also addressed this. After talking with a handful of folks, I don't think this fix is correct. https://github.com/android/ndk-samples/pull/1000 seems to be the right fix.

PR #952, opened 2023-07-28
Update dependencies used in hello-vulkan.

DanAlbert:


DanAlbert:
Hmm, that build failure is annoying:

```
FAILURE: Build failed with an exception.

* Where:
Build file '/home/runner/work/ndk-samples/ndk-samples/hello-vulkan/app/build.gradle' line: 17

* What went wrong:
A problem occurred evaluating project ':app'.
> Failed to apply plugin 'com.android.internal.application'.
   > Android Gradle plugin requires Java 17 to run. You are currently using Java 11.
      Your current JDK is located in /opt/hostedtoolcache/jdk/11.0.20/x[64](https://github.com/android/ndk-samples/actions/runs/5695821544/job/15439743773?pr=952#step:7:65)
      You can try some of the following options:
       - changing the IDE settings.
       - changing the JAVA_HOME environment variable.
       - changing `org.gradle.java.home` in `gradle.properties`.
```

I changed it to 17 for the same error I had locally. Might be that we need to update the JDK used on the buildbot to match the one used by newer versions of Studio?

fmayer:
Verified it works by using `__ANDROID_API__` and `#error` in the file.

DanAlbert:
(I know the "new" NDK here is now old, but it's an old PR)

enh-google (approved):

PR #955, opened 2023-08-03
Remove fragile gradle parsing.

DanAlbert:
My PGO sample broke this because it uses double quotes. Rather than fix the "parser", just explicitly list the versions we use here. Auto-detection is nice if we *want* to have a lot of versions supported, but we don't.


DanAlbert:
https://github.com/android/ndk-samples/issues/1010 will make this unnecessary.

PR #969, opened 2023-12-14
Sanitizer: Set useLegacyPackaging to true

robotchaoX:
[Why]
Property useLegacyPackaging determin whether to use the legacy convention of compressing all .so files in the APK. If null, .so files will be uncompressed and page-aligned when minSdk >= 23. Otherwise, it failed with "INSTALL_FAILED_INVALID_APK: Failed to extract native libraries, res=-2" when minSdk >= 23.

[How]
Set useLegacyPackaging to true when miniSdk >=23.
Reference: https://developer.android.com/ndk/guides/asan


robotchaoX:
Hi @jfgoog ,
Could you review this commit? Thanks!

PR #970, opened 2023-12-14
Sanitizer: Set useLegacyPackaging for asan

robotchaoX:
[Why]
Property useLegacyPackaging determin whether to use the legacy convention of compressing all .so files in the APK. If null, .so files will be uncompressed and page-aligned when minSdk >= 23. Otherwise, it failed with "INSTALL_FAILED_INVALID_APK: Failed to extract native libraries, res=-2" when minSdk >= 23.
And wrap.sh is only available for API level 27 and above.

[How]
Set useLegacyPackaging to true.
Set compileSdk to 27.

[Reference]
https://developer.android.com/ndk/guides/asan

### DIFF
--- a/sanitizers/app/build.gradle
+++ b/sanitizers/app/build.gradle
@@ -37,6 +37,12 @@ android {
         asan {
             initWith debug
             debuggable true
+            compileSdk 27
+            packagingOptions {
+                jniLibs {
+                    useLegacyPackaging true
+                }
+            }
             externalNativeBuild {
                 cmake {
                     arguments "-DANDROID_ARM_MODE=arm", "-DANDROID_STL=c++_shared", "-DSANITIZE=asan"
--- a/sanitizers/app/build.gradle
+++ b/sanitizers/app/build.gradle
@@ -10,6 +10,8 @@ android {
 
     defaultConfig {
         applicationId "com.example.sanitizers"
+        // If you raise minSdk to 23 or higher, make sure you've read the note
+        // below about useLegacyPackaging.
         minSdk 21
         targetSdk 32
         versionCode 1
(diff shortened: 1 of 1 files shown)


sanitizers/app/build.gradle
```diff
@@ -37,6 +37,12 @@ android {
         asan {
             initWith debug
             debuggable true
+            compileSdk 27
```
fmayer, line 40: compileSdk is already 32 above, and you really want to change minSdk. But it's probably fine to just leave this out.

robotchaoX: Solved it. Thanks!

fmayer (approved):

fmayer:
@DanAlbert @enh-google PTAL

enh-google (approved):
(this one i think i do understand, and it does match what i believe to be true. but you might still want to wait for danalbert next week if you want more than a "seems legit?" from someone who doesn't really know what they're talking about :-) )

DanAlbert:
I added explanatory comments but otherwise LGTM.

DanAlbert:
... but our CI's broken. I'll go look into that.

DanAlbert:
GitHub is currently in the middle of rolling out changes to the GHA images that alter how the Android tools are installed: https://github.com/actions/runner-images/issues/8952. That bug says that `sdkmanager` is supposed to still be there, but maybe that isn't happening. I see Sentry updated their CI to use https://github.com/android-actions/setup-android, which I think tells us the problem:

> Adding platform-tools (contains adb) and cmdline-tools/11.0/bin (contains sdkmanager) to $PATH.

I guess `sdkmanager` moved to a non-predictable location? Awesome. I'll send a PR to switch us over to setup-android.

DanAlbert:
https://github.com/android/ndk-samples/pull/975 appears to fix it. Once that merges I'll rebase this PR so the checks can re-run, then it should be good to merge.

PR #971, opened 2023-12-14
Sanitizer: Make it more portable

robotchaoX:
[Why]
Use cmake variable rather than fixed name sanitizers. Make it more portable and has no impact.

[How]
Use cmake variable rather than fixed name sanitizers.


DanAlbert:
I don't understand what this makes more portable. Did you see the build.gradle? This _can't_ be fully driven from CMakeLists.txt, as some of the critical decisions are already made by the time CMakeLists.txt is "run".

robotchaoX:
Thanks for your review. Sorry for the immature idea. Drop this PR.

PR #983, opened 2024-02-26
Move member functions to a cpp file in hello-vulkan sample

kocdemir:
This PR moves class functions into a cpp file to make improve the readability.


kocdemir:
@DanAlbert I've separated the previous PR into 2 commits in this one to ease the review. Thanks

kocdemir:
Sure, this is not urgent and not changing the behavior much. Original owner of the sample, @GOOG-sergiu, can help with reviewing Vulkan related changes. Thank you!

PR #994, opened 2024-04-06
Support for armv8 arch only devices

nfynt:
Recent pixel devices do not report armv7 arch support and only arm64-v8a.  This causes benchmark test to exit early with message stating "Not an ARMv7 and not an X86 SSSE3 CPU !\n".

This update includes check for `ANDROID_CPU_FAMILY_ARM64` along with subsequent cmake and gradle updates for running the test app.


hello-neon/app/src/main/cpp/helloneon.c
```diff
@@ -114,7 +114,7 @@ jstring Java_com_example_helloneon_HelloNeon_stringFromJNI(JNIEnv* env,
   strlcat(buffer, "Neon version   : ", sizeof buffer);
 
   family = android_getCpuFamily();
-  if ((family != ANDROID_CPU_FAMILY_ARM) &&
+  if ((family != ANDROID_CPU_FAMILY_ARM) && (family!=ANDROID_CPU_FAMILY_ARM64) &&
```
DanAlbert, line 117: I'm pretty sure this also works for x86_64, and rather than continuing to extend this list, you should instead just remove this block. I'm not sure what rv64 plans to do about this, but I'll talk with them whenever the time comes for that.

DanAlbert (changes requested):
There's actually quite a lot here that I think can be safely deleted, since Neon is available by default for a few NDK releases now. I'm not totally sure how much I can purge though, so I'll merge this and take a look at the cleanup separately.

For that matter, this whole sample would probably be more useful as a generic vectorization sample rather than specific to Neon, since there are also clang's arch-generic vector APIs, as well as SVE coming up with newer versions of armv8, and whatever rv64 has planned, but that's definitely out of scope for this PR :)

DanAlbert (approved):

DanAlbert:
I think you probably need to update https://github.com/android/ndk-samples/blob/ece52f26503987bbfd2ded8c73ffde05c457d035/hello-neon/app/src/main/cpp/helloneon-intrinsics.c#L18?

DanAlbert:
Actually it was the CMakeLists.txt that needed updating. Pushed the fix to your branch.

PR #998, opened 2024-04-16
fix native-activity ClassNotFoundException

heyoukang:
fix issue https://github.com/android/ndk-samples/issues/929

why?
Library 'androidx.appcompat:appcompat:1.5.1' merged 'InitializationProvider' to final AndroidManifest.xml，and it also add InitializationProvider.class to final apk.
but android:hasCode="false" make this apk is pure native apk, so InitializationProvider.class doesn't work. we can clearly see the crash info DexPathList is empty.

Didn't find class "androidx.startup.InitializationProvider" on path: DexPathList[[],nativeLibraryDirectories

so AndroidManifest.xml has InitializationProvider info, but dex doesn't. that's why we get ClassNotFoundException.

there are two solutions.

remove useless library 'androidx.appcompat:appcompat:1.5.1', Neither AndroidManifest.xml nor dex have InitializationProvider info
remove android:hasCode="false", make dex work well
this pr remove useless library, and also remove android:hasCode="false" to avoid someone stuck in this problem again.
this is just sample, so I think it's ok.

### DIFF
--- a/native-activity/app/build.gradle
+++ b/native-activity/app/build.gradle
@@ -33,6 +33,4 @@ android {
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
-    implementation 'androidx.appcompat:appcompat:1.5.1'
-    implementation 'androidx.constraintlayout:constraintlayout:2.1.4'
 }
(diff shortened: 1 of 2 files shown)


heyoukang:
@DanAlbert Can you help me merge this pr?

heyoukang (commented):
ok

DanAlbert:
https://github.com/android/ndk-samples/pull/1000 is pretty much the same but includes docs and purges the jar dependencies (of which there are none currently, but it's just another booby trap waiting for someone). Closing this in favor of that PR. Thanks for the help in figuring it out 👍

PR #1000, opened 2024-04-16
Remove unused dependencies from native-activity.

DanAlbert:
These dependencies were not actually used but were in contradiction with the `android:hasCode=false` in the manifest. Remove the pointless dependencies, but leave some comments explaining to anyone that copies from here that they need to update `app:hasCode` if they add any DEX to their app.

Fixes https://github.com/android/ndk-samples/issues/929.


native-activity/app/src/main/AndroidManifest.xml
```diff
@@ -4,7 +4,13 @@
     android:versionCode="1"
           android:versionName="1.0">
 
-  <!-- This .apk has no Java code itself, so set hasCode to false. -->
+  <!--
+  This .apk has no Java/Kotlin code, so set hasCode to false.
```
enh-google, line 8: i'm assuming you _have_ to set this to `false` for this to work at all? is it worth saying _why_?

enh-google (approved):

native-activity/app/src/main/AndroidManifest.xml
heyoukang: 截屏2024-04-17 上午9 53 41
hi DanAlbert and enh-google, I hold the different view.
set hasCode to false just for bytecode optimized, it is not ‘must have’, and it also can be a trap when someone want to add Jave code. especially for engineers who are not app developers. 

explain why this error: https://github.com/android/ndk-samples/pull/998

DanAlbert: > i'm assuming you have to set this to false for this to work at all? is it worth saying why?

My best guess is that it improves app startup time, but I'm still waiting to hear from anyone that can say with certainty what it does. hackbod wrote the original comment rather than removing the property, so I assume it is worth keeping.

DanAlbert: (resolving so I can merge since this is definitely not worse, I'll be back with another PR if I learn that it should be removed)

DanAlbert: No, it should not be removed. Setting it incorrectly has no symptoms on the device I tested with, but it may on others, and may in the future (which is how this problem cropped up in the first place, the bug was asymptomatic for a time and then wasn't, we shouldn't repeat that in the other direction): https://developer.android.com/guide/topics/manifest/application-element#code

heyoukang: I don't think it would has symptoms on the other devices or in the future, but it definitely not worse. so I agree with you. 🍺 🍻

Issue #999, opened 2024-04-16
Replace `ALooper_pollAll` with `ALooper_pollOnce`


DanAlbert:
`ALooper_pollAll` has been (soft) removed from the NDK because the implementation has a race so it can't be used safely. `ALooper_pollOnce` should be used instead.

DanAlbert:
Fixed by https://github.com/android/ndk-samples/pull/1008 and https://github.com/android/ndk-samples/pull/1009

PR #1002, opened 2024-04-24
Replace `ALooper_pollAll` with `ALooper_pollOnce`.

DanAlbert:
The `pollAll` API was removed because it can't be used safely. The recommendation is to replace existing calls to `pollAll` with `pollOnce` with an explicit loop.

There are quite a few other samples that are doing this. Fixing one to start so I can check with someone that understands the `pollAll` bug so I don't accidentally copy the same bug, then I'll fix the others.

https://github.com/android/ndk-samples/issues/999

### DIFF
--- a/native-activity/app/src/main/cpp/main.cpp
+++ b/native-activity/app/src/main/cpp/main.cpp
@@ -27,13 +27,18 @@
 #include <cerrno>
 #include <cstdlib>
 #include <cstring>
+#include <functional>
 #include <initializer_list>
 #include <memory>
 
-#define LOGI(...) \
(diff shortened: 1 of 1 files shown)


DanAlbert:
@jreck says this isn't actually a good fix. The problem is present when calling `pollOnce` in a loop as I have here as well. The nuance is that this sample doesn't have the bug at all, because the problem only happens when `ALooper_wake` is used, and this sample doesn't do that. OTOH, this sample doesn't do that mainly because it's trivial. Anything built with this sample as its base likely would.

The better fix is to migrate to `AChoreographer`. I'll close this and upload a new PR that does that.

PR #1004, opened 2024-04-30
Move sensor handling to a callback.

DanAlbert:
The apparent change in `minSdkVersion` here isn't actually changing much. The NDK hasn't supported ICS in a long time, and 21 is the minimum for modern NDKs (though I think this sample is still using an NDK that supports 19, it won't soon). The raise is needed for `android_set_abort_message`.

### DIFF
--- a/native-activity/app/src/main/cpp/main.cpp
+++ b/native-activity/app/src/main/cpp/main.cpp
@@ -30,10 +30,13 @@
 #include <initializer_list>
 #include <memory>
 
-#define LOGI(...) \
-  ((void)__android_log_print(ANDROID_LOG_INFO, "native-activity", __VA_ARGS__))
-#define LOGW(...) \
-  ((void)__android_log_print(ANDROID_LOG_WARN, "native-activity", __VA_ARGS__))
+#define LOG_TAG "native-activity"
+
+#define _LOG(priority, fmt, ...) \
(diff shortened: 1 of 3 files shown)


native-activity/app/src/main/cpp/main.cpp
```diff
@@ -35,9 +36,34 @@
...
 #define LOGW(fmt, ...) _LOG(ANDROID_LOG_WARN, (fmt)__VA_OPT__(, ) __VA_ARGS__)
 #define LOGI(fmt, ...) _LOG(ANDROID_LOG_INFO, (fmt)__VA_OPT__(, ) __VA_ARGS__)
 
+[[noreturn]] __attribute__((__format__(__printf__, 1, 2))) static void fatal(
+    const char* fmt, ...) {
+  va_list ap;
+  va_start(ap, fmt);
+  char buf[BUFSIZ];
+  if (vsnprintf(buf, sizeof(buf), fmt, ap) < 0) {
+    android_set_abort_message("failed for format error message");
```
enh-google, line 49: vasprintf()? you're not assuming heap corruption here, right?

enh-google (approved):

native-activity/app/src/main/cpp/main.cpp
DanAlbert: No, I just always forget about that variant.

enh-google (approved):

PR #1008, opened 2024-05-02
Replace ALooper_pollAll with ALooper_pollOnce.

DanAlbert:
ALooper_pollAll can't be called safely and is being removed from the NDK. Replace with ALooper_pollOnce.

### DIFF
--- a/native-activity/app/src/main/cpp/main.cpp
+++ b/native-activity/app/src/main/cpp/main.cpp
@@ -404,25 +404,21 @@ void android_main(android_app* state) {
     engine.state = *(SavedState*)state->savedState;
   }
 
-  while (true) {
-    // Read all pending events.
-    int events;
-    android_poll_source* source;
-
+  while (!state->destroyRequested) {
     // Our input, sensor, and update/render logic is all driven by callbacks, so
     // we don't need to use the non-blocking poll.
-    while ((ALooper_pollAll(-1, nullptr, &events, (void**)&source)) >= 0) {
-      // Process this event.
-      if (source != nullptr) {
-        source->process(state, source);
-      }
+    android_poll_source* source = nullptr;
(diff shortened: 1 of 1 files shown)


native-activity/app/src/main/cpp/main.cpp
```diff
@@ -404,25 +404,28 @@ void android_main(android_app* state) {
...
-      if (source != nullptr) {
-        source->process(state, source);
-      }
-
-      // Check if we are exiting.
-      if (state->destroyRequested != 0) {
-        engine_term_display(&engine);
-        return;
-      }
+    android_poll_source* source;
```
jreck, line 410: initialize to null

native-activity/app/src/main/cpp/main.cpp
```diff
@@ -404,25 +404,28 @@ void android_main(android_app* state) {
...
+    auto result = ALooper_pollOnce(-1, nullptr, nullptr,
+                                   reinterpret_cast<void**>(&source));
+    switch (result) {
```
jreck, line 413: this is fine but couldn't this just be an `if result == error { die() } else if (source) { source->process }` ? like the other return values are ignored, so a switch is kinda overkill?

DanAlbert: I wasn't clear on how `source` behaved for the other return values. If that's safe, works for me.

DanAlbert: Oh, and I'd avoided the fatal error because the docs don't say what the failure cases are, so I'm not sure how critical those actually are. Dug through the implementation and history and it looks like they're all unrecoverable (`epoll_wait` failed, or the thread has no looper, and afaict I can't create one if it's missing), so yeah, fatal. https://r.android.com/c/3073883 expands the docs.

jreck (approved):

PR #1013, opened 2024-05-06
Run upgrade assistant on all samples.

DanAlbert:
Not quite all. I skipped the other-builds directory because 'tis a silly place that needs to just be merged with the core samples properly as a build variant rather than being a distinct project.


jfgoog (approved):
I did not scrutinize every diff; I assume these are all automated changes. Anything you did that was manual should be in a separate change.

PR #1023, opened 2024-05-07
Remove all ndk-build samples.

DanAlbert:
None of these show anything interesting. At first glance I thought a few might be worth preserving, but the few interesting things they do are actually superfluous. I'm removing all of them for now and will add a new sample soon specifically to show how to use ndk-build in a substantial enough way to be useful.


jfgoog (approved):

DanAlbert:
Okay, none of these are actually interesting. The most they manage to do is show how to do the gradle side of this, but the `Android.mk` themselves are badly written because they're trying to share with their "parent" samples. I'll add a focused sample for ndk-build soonish.

PR #1026, opened 2024-05-07
Add top level samples project.

DanAlbert:
This is step one of https://github.com/android/ndk-samples/issues/1010. It's not really finished until I've removed all the stuff left behind (which I'm leaving for another PR because it's just a big mess of deletes that'd make this PR harder to read).

It's may be good to also remove the mostly unnecessary middle directories from most of these. Doing that will break a bunch of existing URLs, but if the goal is to be a project that's actually sensible to copy, we need to rearrange to do that. Maybe that's better left for a broader rearranging that's probably overdue anyway, or maybe we should just leave it as-is to keep the URLs alive.


settings.gradle
```diff
@@ -0,0 +1,65 @@
...
+include(":hello-libs:app")
+// To generate libs used in this sample:
+//   1) enable the gen-libs at end of this file
+//   2) enable module build dependency in app/build.gradle
+//   3) build the app's APK in Android Studio or on command line
+//   4) undo step 1) and 2) above
+// include(":hello-libs:gen-libs")
+include(":hello-neon:app")
+include(":hello-oboe:app")
+//include(":hello-vulkan:app")
```
DanAlbert, line 43: Oops, I forgot to come back to this.

DanAlbert: https://github.com/android/ndk-samples/pull/1027 fixes the problem with the sample.

jfgoog (approved):

PR #1028, opened 2024-05-08
Remove unindexed packaging.yaml files.

DanAlbert:
The service that reads these now only reads the top level one.


DanAlbert:
DO NOT MERGE (yet): the service hasn't updated yet.

jfgoog (approved):

PR #1029, opened 2024-05-08
Remove unused gradle stuff in child projects.

DanAlbert:
None of these are used now that the top level one exists.

https://github.com/android/ndk-samples/issues/1010


DanAlbert:
Do not merge until the samples index service has updated.

DanAlbert:
Index has been updated, this is safe to merge now.

PR #1035, opened 2024-05-09
Migrate the rest of the samples to the convention plugin.

DanAlbert:
All the module build.gradle files are in a pretty good state after this. This is the same sort of stuff that was done to audio-echo: migrate to the convention plugin, remove unused crap, make whatever was left more presentable.

Depends on https://github.com/android/ndk-samples/pull/1034.


jfgoog (approved):
So much duplication and boilerplate eliminated. I like it.

PR #1036, opened 2024-05-09
Move the fast part of the "build" sooner.

DanAlbert:
I'm tired of waiting for the build to finish only to learn that my markdown isn't markdowny enough.


DanAlbert:
(it'd probably be even better for this to be its own check, but I don't want to dig into redoing that right now, and probably YAGNI)

jfgoog (approved):

Issue #1031, opened 2024-05-08
camera/basic renders black screen on API 34


DanAlbert:
Reverting all the way to f50ef3c07fcb5c9240daf47280c348eb31a158a4 doesn't fix it, so it's not a result of my recent changes. The sample was working for me previously, and I think the only thing I've changed is that I've flashed my 4a 5G forward from API 32 (I think that's what it was, maybe 31) to API 34.

DanAlbert:
Same behavior on an API 34 emulator. It _does_ work on an API 32 emulator. It's also broken on API 33, so something changed between 32 and 33.

PR #1047, opened 2024-05-13
camera:basic: fail if permissions are not granted

DanAlbert:
Pre-work for https://github.com/android/ndk-samples/issues/1031. For some reason the UI for the storage permission isn't shown whenever targetSdkVersion is 33+, and it's just returning a denial. This sample is badly structured such that if either permission isn't granted, the UI init will never run. The two permissions requests really ought to be decoupled (the viewfinder should still work even if I never try to save a photo), and the UI should tell the user why it failed. For now I'm just making it fail more violently, because the lack of any error or logging is what made this a multi-hour debugging experience instead of the now obvious problem.

### DIFF
--- a/camera/basic/src/main/java/com/sample/camera/basic/CameraActivity.java
+++ b/camera/basic/src/main/java/com/sample/camera/basic/CameraActivity.java
@@ -41,6 +41,8 @@ import android.widget.Toast;
 
 import static android.hardware.camera2.CameraMetadata.LENS_FACING_BACK;
 
+import java.util.Arrays;
+
 class CameraSeekBar {
     int _progress;
     long _min, _max, _absVal;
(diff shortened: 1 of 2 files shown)


camera/basic/src/main/java/com/sample/camera/basic/CameraActivity.java
```diff
@@ -212,19 +214,53 @@ public void RequestCamera() {
...
-            notifyCameraPermission(grantResults[0] == PackageManager.PERMISSION_GRANTED &&
-                    grantResults[1] == PackageManager.PERMISSION_GRANTED);
+        if (permissions.length == 0) {
+            // https://developer.android.com/reference/androidx/core/app/ActivityCompat.OnRequestPermissionsResultCallback#onRequestPermissionsResult(int,java.lang.String[],int[])
+            //
+            // Note: It is possible that the permissions request interaction with the user is
+            // interrupted. In this case you will receive empty permissions and results arrays which
+            // should be treated as a cancellation.
+            //
+            // The docs aren't clear about *why* is might be canceled, so it's not clear what we
```
enh-google, line 232: s/is/it/

enh-google: can you find any 1p app code and see what they're doing? (and/or file a bug against the docs.)

enh-google (approved):

camera/basic/src/main/java/com/sample/camera/basic/CameraActivity.java
DanAlbert: Yeah, hadn't quite gotten to filing a bug yet but that was the plan. http://b/340568363

PR #1061, opened 2024-05-15
Add absl.

DanAlbert:
absl doesn't provide an AAR, and the AGP DSL for creating one can't describe the package correctly (it can't seem to find the libraries at all, and even if it could, it can't describe a single header directory shared by multiple libraries). Without that support, the only remaining option is to vendor it into the tree and include it directly in any CMake project that uses it. The downside of doing it this was is that absl will be rebuilt for every module rather than shared among them.

An alternative would be importing android-base for AOSP, but that's a rather large porting task.

Another alternative would be doing none of this and writing our own, much simpler subsets of those libraries. Mostly we want `CHECK`, `LOG`, and `DISALLOW_COPY_AND_ASSIGN`. libbase's logging.cpp has quite a few dependencies on other parts of libbase (file, parseint, strings, threads) and it quickly becomes a large project to import, but maybe those are only needed by APIs we don't actually care about.


DanAlbert:
Closing in favor of https://github.com/android/ndk-samples/pull/1062.

PR #1062, opened 2024-05-15
Import a subset of libbase, build an AAR.

DanAlbert:
It's still quite a lot for just CHECK, LOG, and
DISALLOW_COPY_AND_ASSIGN. The strings stuff is only needed because r26 is too old to have the stdlib string split/starts_with/etc. Maybe I'll end up wanting more later and this is a worthwhile start?


DanAlbert:
I was able to excise all the stuff I wasn't using, and have made pretty extensive use of this in two draft samples I have, so either this or https://github.com/android/ndk-samples/pull/1061 ought to be merged. The alternative "write your own" at this point wouldn't be much smaller than this is, but would be worse.

enh-google (approved):
this seems less unmanageable than the alternative...

DanAlbert:
> this seems less unmanageable than the alternative...

Both this and absl seem about as easy to maintain to me, but actually a better reason to go this route than the absl route: I need this in libraries. The lack of ABI stability wouldn't matter within the samples repo, but I wouldn't want to accidentally mislead someone into thinking it's safe for library use.

enh-google:
> > this seems less unmanageable than the alternative...
> 
> Both this and absl seem about as easy to maintain to me, but actually a better reason to go this route than the absl route: I need this in libraries. The lack of ABI stability wouldn't matter within the samples repo, but I wouldn't want to accidentally mislead someone into thinking it's safe for library use.

well, that's part of what i mean --- not wanting to be on the hook for absl updates (and breaking source/abi compatibility if you do).

Issue #1011, opened 2024-05-03
rewrite hello-neon as a generic vectorization sample


DanAlbert:
This sample probably made more sense back when it was written, but now it's not really guiding people toward doing the right thing.

Much more useful, I think would be rewriting it as a general purpose vectorization demo. Pick some sort of vectorization problem (probably something simple like a matrix multiply), and build an implementation for each of:

* Neon intrinsics
* Clang's arch-generic intrinsics
* SVE
* C++'s `<algorithm>` stuff, if the problem we pick is suitable for them
* Plain old C++, letting the compiler autovectorize

And then the UI would show the benchmark results of each so people can see the differences.

DanAlbert:
After this merges and before I delete the old one, https://developer.android.com/ndk/guides/cpu-arm-neon will need to be updated since it points to hello-neon.

rishuk58:
.google/packaging.yaml

PR #1067, opened 2024-06-12
Add vectorization sample.

DanAlbert:
This adds a new generic vectorization sample to replace hello-neon. Most importantly, it covers non-Neon options for SIMD. One of the useful things it shows, for example, is that there's actually no reason to write SIMD code the way that hello-neon does any more.

This also solves a much simpler problem (small matrix multiplication), which makes it easier to see how to deal with the SIMD features rather than figuring out what a FIR filter is.

Finally, this sample benchmarks each of the implementations so it's obvious what is and isn't worth doing. I was sort of surprised that auto-vectorization didn't do better, and was pleased to learn that there's no reason at all to write Neon intrinsics.

I'll delete hello-neon after this merges and I've fixed up the doc links.

https://github.com/android/ndk-samples/issues/1011


DanAlbert:
Currently a draft because I seem to have encountered an AGP bug that makes cross-module dependencies flaky, so I may need to deal with `base` in a different way...

vectorization/README.md
```diff
@@ -0,0 +1,149 @@
...
+need to either alter the algorithm to tile the operations, or use Scalable
+Vector Extensions (AKA [SVE]).
+
+However, the benefit of the portability trade-off is that this does outperform
+the auto-vectorization implementation.
+
+See
+https://clang.llvm.org/docs/LanguageExtensions.html#vectors-and-extended-vectors.
+
+### Neon
```
DanAlbert, line 69: Should I maybe delete the code backing this and move this explanation to the "not shown" section, or is it worthwhile to have the "look, proof that this is pointless" example?

vectorization/README.md
```diff
@@ -0,0 +1,149 @@
+# Vectorization
+
+This sample shows how to implement matrix multiplication using various
+vectorization approaches.
+
+Note: You should not reuse this matrix library in your application. It was not
+written to be useful beyond the scope of this demo. If you're looking for a
+matrix library, you probably want [GLM] for graphics applications, or a linear
+algebra library for compute applications.
```
enh-google, line 9: "such as BLAS"?

vectorization/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+Given the relatively small problem size used here (4x4 matrices and vec4s), the
+best performing implementations in this sample are the ones that can best
+improve over the naive implementation without large set up costs. You should not
+take the results of this sample as authoritative: if performance is important to
+you, you **must** benchmark your code for workloads realistic for your app.
+
+If you're not familiar with it [Godbolt] is an invaluable tool for examining
+compiler optimizer behavior. You could also use `llvm-objdump` from the command
+line, but that's just Godbolt with extra steps.
```
enh-google, line 24: i usually use `cc -S` for "local godbolt". objdump tends to solve the related but different problem of "i can't actually reduce this easily --- let's look at the real code in situ".

enh-google: (strictly, `cc -S -o - -O2`, but ykwim)

vectorization/README.md
```diff
@@ -0,0 +1,149 @@
...
+### Clang vectors
+
+See [clang_vector.h] for the implementation.
+
+This implementation uses Clang's generic vector types. This code is mostly as
+portable as the auto-vectorization implementation, with the only caveat being
+that it is limited by the width of the vector registers for the target hardware.
+To deal with problems that don't fit in the target's vector registers, you would
+need to either alter the algorithm to tile the operations, or use Scalable
+Vector Extensions (AKA [SVE]).
```
enh-google, line 61: strange to mention this at all if you're not going to talk about it in more detail (or even explain _why_ you're not talking about it more).

vectorization/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+See
+https://clang.llvm.org/docs/LanguageExtensions.html#vectors-and-extended-vectors.
+
+### Neon
+
+See [neon.h] for the implementation.
+
+This implementation uses the Arm-specific Neon vector types. The implementation
+is nearly identical to the one in [clang_vector.h]. The only differences how the
```
enh-google, line 74: "are how"

vectorization/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Alternatives not shown here
+
+There are other approaches that could be used that aren't shown here.
+
+[SVE] scales SIMD to arbitrarily sized vectors, and the C extensions, while
+making for less concise code than is needed for a constrained vector size like
+we have here, handle windowing of data to fit the hardware vector size for you.
+For problems like the small matrix multiply we do here, it's overkill. For
+portability to a wide variety of (Arm) CPUs, it can reduce the difficulty of
```
enh-google, line 130: this makes it sound more available than it is ... "qualcomm has not shipped SVE" is a pretty huge caveat for real-world use.

vectorization/src/main/cpp/auto_vectorization.h
```diff
@@ -0,0 +1,56 @@
...
+ * @param lhs The left operand.
+ * @param rhs The right operand.
+ * @return The result of lhs * rhs.
+ */
+template <typename T, size_t M, size_t N, size_t P>
+Matrix<M, P, T> MultiplyWithAutoVectorization(const Matrix<M, N, T>& lhs,
+                                              const Matrix<N, P, T>& rhs) {
+  // This may look like an unfair benchmark because this implementation uses the
+  // less vector friendly one than the others, however, using the vector
+  // friendly algorithm here actually made performance worse.
```
enh-google, line 42: explicitly remind people that this is why it's so important to test their actual implementation, and not just rely on what the internet told them.

it's probably also worth saying somewhere that this is one reason why the professionals (the codec authors etc) don't rely on stuff like autovectorization --- because you can't _rely_ on it. you're at the mercy of today's version of the compiler.

vectorization/src/main/cpp/benchmark.cpp
```diff
@@ -0,0 +1,127 @@
...
+  }};
+
+  switch (backend) {
+    case Backend::kAutoVectorization:
+      LOG(INFO) << "Benchmarking auto-vectorization";
+      return Benchmark(position, translation, [](Vec4 p, Mat4 t) {
+        return MultiplyWithAutoVectorization(t, p);
+      });
+    case Backend::kCxxSimd:
+#if __NDK_MAJOR__ >= 28
```
enh-google, line 90: speaking of ndk versions, do we have FMV in the ndk yet? might be worth (in the section about "test your actual code on your actual use case") reminding people that also doesn't relieve you of the need to actually benchmark --- you can't just assume that what it generates will be optimal for each variant, nor that it's always obvious which variant is best for the hardware it's running on.

vectorization/src/main/cpp/matrix.h
```diff
@@ -0,0 +1,148 @@
...
+ private:
+  std::array<T, Rows * Columns> cells_ = {};
+};
+
+// Enables automatic deduction of definitions like `Matrix m{{1, 0}, {0, 1}}`
+// without needing to specify `Matrix<2, 2, int>`.
+template <size_t Rows, size_t Columns, typename T>
+Matrix(T (&&)[Rows][Columns]) -> Matrix<Rows, Columns, T>;
+
+// TODO: Once we move to NDK r28, these can `template <typename T = float>`.
```
enh-google, line 144: #if like above?

enh-google (approved):

vectorization/README.md
DanAlbert: I only know a couple by name and wasn't sure if any of them were actually good :)

vectorization/README.md
DanAlbert: It's mentioned below.

vectorization/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Clang vectors
+
+See [clang_vector.h] for the implementation.
+
+This implementation uses Clang's generic vector types. This code is mostly as
+portable as the auto-vectorization implementation, with the only caveat being
+that it is limited by the width of the vector registers for the target hardware.
+To deal with problems that don't fit in the target's vector registers, you would
+need to either alter the algorithm to tile the operations, or use Scalable
```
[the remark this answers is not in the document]
DanAlbert: It is linked? Is there a better doc I should be linking to? I didn't find much.

vectorization/src/main/cpp/benchmark.cpp
DanAlbert: Reading a bit more about this, there's more than one form of it: https://extensa.tech/blog/migrating-multiversioning/

We have the "old" kind for sure, which I can mention but I don't have enough hardware range to test more than 2 variations (though one of those is denver and one of those is armv8.2, so it's enough of a spread to potentially be interesting), but that's the kind that's meant more for *writing* different implementations for each target, which I wouldn't do here.

The new kind tells me "Attribute 'target_clones' multiversioned functions do not yet support function templates", o it doesn't work here. I'll mention it in the README though, since it's useful for someone that didn't pick a problem that needs templates :)

vectorization/README.md
DanAlbert: That's fair. I've toned it down, but avoided being specific because I know we won't remember to come fix it as the hardware does roll out. PTAL

DanAlbert:
Added a fair amount of content to the readme for explaining FMV. That probably belongs on DAC rather than here, but it can go here for now.

vectorization/README.md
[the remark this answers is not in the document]
enh-google: "tile the operations" isn't a phrase in my vocabulary.

vectorization/README.md
```diff
@@ -0,0 +1,182 @@
...
+Vector Extensions (AKA [SVE]).
```
enh-google, line 61: does SVE actually help with this problem? (have you considered that negge might be a better reviewer for this stuff? :-) )

vectorization/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+A Neon implementation would be nearly identical to the one in [clang_vector.h].
+The only difference is how the vector type is specified. A lot of older Neon
+sample code looks substantially different because it uses the Neon intrinsics
+defined in `arm_neon.h`, but if you look at how the intrinsics in that file are
+defined, all they actually do (for a little endian system, and Android does not
+support big endian, so we can ignore that caveat) is use the `*` operator and
+leave the correct instruction selection up to Clang.
+
+In other words, you should probably never use the Neon-specific approach. The
```
enh-google, line 112: (this really seems worth a separate note on the d.a.c neon page...)

vectorization/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+There are two compiler attributes that can be helpful for targeting specific
+hardware features when optimizing hot code paths: [target] and [target_clones],
+both of which may be referred to as "function multiversioning" or "FMV". Each
+solves a slightly different but related problem.
+
+The `target` attribute makes it easier to write multiple implementations for a
+function that should be selected based on the runtime hardware. If benchmarking
+shows that one implementation performs better on armv8.2 and a different
+implementation performs better on armv8 (see the docs for more details on
```
enh-google, line 147: maybe explicitly say 8.0 or something?

enh-google (approved):

vectorization/README.md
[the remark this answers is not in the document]
DanAlbert: Ah, gotcha. I've not actually written or seen this either, but it's what all the references I used said too. I'll see if I can find something that illustrates it better, but aiui it's just a matter of breaking up e.g. a 16x16 matrix multiply into 4 4x4 matrix multiplies.

vectorization/README.md
DanAlbert: I pointed him at it a few days ago but haven't heard back yet. @negge explicitly now that I know his github username :)

vectorization/README.md
DanAlbert: Yes. One thing at a time though :)

vectorization/README.md
DanAlbert: I'd written that originally, but it looked odd enough that I wasn't sure I'd ever actually seen it phrased that way before. SGTM though

vectorization/README.md
DanAlbert: You should still PTAL, but acking and merging for now. It's easy enough to fix docs later.

enh-google: yeah, i'd have been tempted to explicitly mention that the portability to "a wide variety of (Arm) CPUs" only really buys you anything if your code is also going to be run on a server, where you might have a wider vector length. for mobile, your SVE vector length is going to be the same as your neon/simd vector length anyway. (and since the power consumption seems like a laws of physics thing, i'm not expecting that to change.)

but this is still better than what we had.

PR #1076, opened 2024-08-27
Upgrade dependencies.

DanAlbert:


DanAlbert:
appcompat upgrade reverted, since that's the one causing the build break. I'll dig into it later, but it looks like it's maybe incompatible with some other dependency that requires a kotlin stdlib older than the minimum for the new appcompat?

enh-google (approved):

Issue #1060, opened 2024-05-15
merge graphics samples


DanAlbert:
We currently have a lot of opengl samples, most of which don't do anything usefully different from the others:

* gles3jni
* hello-gl2
* san-angeles
* teapots/choreographer-30fps
* teapots/classic-teapot
* teapots/image-decoder
* teapots/more-teapots
* teapots/textured-teapot

All of those could be merged into a single opengl sample that renders a spinning textured cube without losing any of the features shown, and it'd be much easier to understand (from what I've seen in the teapots code, they are not currently good for educating people, even if they already have some opengl experience). The sample would have multiple backends so it could demo the various GL versions. It *might* make sense to merge hello-vulkan into that too.

There's also endless-tunnel, which is a fully fleshed out game. native-activity is technically also GL, but that's not really the point of the sample. Those stay separate.

PR #1079, opened 2024-09-16
Update to CMake 3.30.3.

DanAlbert:


DanAlbert:
The build failure doesn't happen locally :( I'll have to find some time to dig into this a bit, but the project structure in those samples is particularly ugly so this might just wait until I finish replacing those in https://github.com/android/ndk-samples/issues/1060

enh-google (approved):

PR #1097, opened 2025-05-05
Add triggers to .github/workflows/

cartland:
This PR standardizes GitHub Actions triggers in workflow files within `.github/workflows/`.

The goal is to ensure workflows run consistently on `push`, `pull_request`, and `workflow_dispatch` events where appropriate.

This is part of a batch of pull requests across repositories owned by the `android` organization on GitHub.

**Project Owner:** Please review the changes carefully to ensure they are correct and appropriate for this project before approving and merging.

* If you do not think this change is appropriate (e.g., a workflow should NOT run on one of these triggers), please leave a comment explaining why.
* If you think the goal is appropriate but notice a mistake in the implementation, please leave a comment detailing the mistake.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,9 @@
 name: build
 
-on: [push, pull_request]
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
 
 jobs:
   build:
@@ -22,4 +25,3 @@ jobs:
     - name: build samples
       run: |
         ./gradlew build
-


cartland:
@DanAlbert you are correct, this change only makes it possible to manually run the workflow. The original comment is generic and tries to explain various changes in a batch of ~30 changes across the `android` GitHub org, so it's a little misleading here.

From an organization perspective, being able to run GitHub Actions manually is important for long-term maintenance. GitHub Actions results have a default retention period of 90 days, so many of the GitHub Actions results are lost after a period of inactivity. If we do not have a way to run the GitHub Actions from the GitHub UI, it can be difficult to quickly verify that the repository is in a healthy state.

For `ndk-samples` specifically, the value is relatively low as long as someone submits a PR or merge more frequently than the GitHub Actions artifact and log retention period. If possible, we'd like to add `workflow_dispatch` to all repositories so that we don't need to make unnecessary commits in the future just to run GitHub Actions.

Please let me know if you have any additional questions. If you think this is ok to merge, please approve. If you don't think this is appropriate, please share why you don't think we should add it to this repository.

DanAlbert (approved):

DanAlbert:
Thanks for the explanation. Merged.

PR #1101, opened 2025-06-18
Update AppMidiManager.cpp

sensn:
When switching MidiDevices the InputPort needs to be closed first, otherwise the new selected MidiDevice doesn't receive any Messages. This is at least the case on my android Device.

AMidiInputPort_close(sMidiInputPort);

### DIFF
--- a/native-midi/app/src/main/cpp/AppMidiManager.cpp
+++ b/native-midi/app/src/main/cpp/AppMidiManager.cpp
@@ -206,6 +206,7 @@ void Java_com_example_nativemidi_AppMidiManager_startWritingMidi(
  */
 void Java_com_example_nativemidi_AppMidiManager_stopWritingMidi(JNIEnv*,
                                                                 jobject) {
+  AMidiInputPort_close(sMidiInputPort); //Close InputPort to make switching MidiDevices work!
   /*media_status_t status =*/AMidiDevice_release(sNativeSendDevice);
   sNativeSendDevice = NULL;
 }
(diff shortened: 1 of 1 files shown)


DanAlbert (approved):
A lot of the code around this looks suspect to me because there's no error checking, so this may attempt to close a thing that wasn't successfully opened... but that's all pre-existing, and this definitely fixes one of the problems. If the rest of the sample looks anything like these two functions, this sample is probably due for a pretty significant overhaul :(

Thanks for the PR! Will merge, but I can't until you've signed the CLA.

native-midi/app/src/main/cpp/AppMidiManager.cpp
```diff
@@ -206,6 +206,7 @@ void Java_com_example_nativemidi_AppMidiManager_startWritingMidi(
  */
 void Java_com_example_nativemidi_AppMidiManager_stopWritingMidi(JNIEnv*,
                                                                 jobject) {
+  AMidiInputPort_close(sMidiInputPort); //Close InputPort to make switching MidiDevices work!
```
DanAlbert, line 209: Actually, looking a little bit further, you can avoid the problem I mentioned for at least this newly added line by checking that `sMidiInputPort != nullptr` before calling this, and then setting it to `nullptr` after closing.

DanAlbert (changes requested):

native-midi/app/src/main/cpp/AppMidiManager.cpp
sensn: Hello,

I added the nullptr-check. If this sample is updated, it would be great to add BluetoothMidi too. I guess this would be entirly on the Java side. Midi 2.0 would be interessting too.
Best regards,
sensn

PR #1137, opened 2025-09-16
Remove native-media sample.

DanAlbert:
This is essentially an OpenMAX AL duplicate for native-codec. OpenMAX AL isn't deprecated, but it's not recommended either. I'm removing the sample because new code should prefer to use the NDK Media Codec APIs as shown in the native-codec sample.


enh-google:
> > is it worth leaving the README.md there saying what you've just said in the commit message?
> 
> Potentially. I wish I could configure redirects for the entire subdirectory because it's much more likely that anyone following an old link is going to have a link to some file within the sample rather than the readme itself. It bothers me a bit to leave a bunch of stale directories around, but I suppose it'd have some value for anyone wondering where that particular sample went, and we can always delete them later down the line when the deletions themselves are old.

yeah, a redirect would be even better.

i don't suppose leaving a symlink works? (certainly wouldn't work if their link is to a file in the old directory that isn't in the "recommended" directory :-( )

DanAlbert:
A symlink would just open in the file viewer, it wouldn't redirect. It would also make it difficult to clone the repo on Windows, so that's out.

I'll send a PR adding the directories back with a readme as a sign post. It's not perfect, but perfect isn't an option available to me.

Issue #1139, opened 2025-09-16
Merge native-activity and native-plasma


DanAlbert:
These two samples are doing basically the same work with different graphics. The only NDK APIs that native-plasma actually demonstrates are the bare bones of `NativeActivity` and `NativeWindow`; the bulk of the code is for generating the specific effect. The native-activity sample does essentially the same thing with simpler graphics (rotating color clears) using OpenGL and logging the accelerometer data.

We already have a bunch of OpenGL samples, and the point of native-activity is (as far as I can tell) to be essentially a `NativeActivity` hello world. I think it'll do that better if we strip out the sensor and GL code, but keep the trivial color clearing, just use the `NativeWindow` APIs for that instead. Once that's done, native-plasma no longer does anything remarkable that native-window doesn't, so it can just be deleted.

PR #1140, opened 2025-09-17
Simplify native-activity, remove now duplicate native-plasma.

DanAlbert:
We don't need OpenGL just to do a color fill. Save that for the OpenGL samples. Replace that with a trivial AHardwareBuffer based color fill.

I've altered the old animation which didn't really work the way I think it was intended to. It looks like someone had intended for the accelerometer to alter the color, but it actually only animated between black and bright green. I've removed the sensor code as superfluous anyway, so just do something simpler: switch from red, to green, to blue every second. I kept some amount of animation rather than just clearing a solid color because if we only draw once and have no update cycle then choreographer and most of the event handling code also becomes useless, and without that the sample no longer bears any resemblance to a real app.

Fixes https://github.com/android/ndk-samples/issues/1139.


native-activity/app/src/main/cpp/main.cpp
```diff
@@ -138,169 +145,68 @@ struct Engine {
...
-    eglTerminate(engine->display);
+
+    ANativeWindow_unlockAndPost(app->window);
   }
-  engine->Pause();
-  engine->display = EGL_NO_DISPLAY;
-  engine->context = EGL_NO_CONTEXT;
-  engine->surface = EGL_NO_SURFACE;
-}
+};
```
enh-google, line 197: stray `;`?

native-activity/app/src/main/cpp/main.cpp
```diff
@@ -319,18 +225,12 @@ static void engine_handle_cmd(android_app* app, int32_t cmd) {
  * event loop for receiving input events and doing other things.
  */
 void android_main(android_app* state) {
-  Engine engine{};
+  Engine engine;
```
enh-google, line 228: neither {} nor memset()?

(did you mean to do `Engine engine = { .app = state };` but stop part way through?)

DanAlbert: Reworked this to be an actual constructor. The behavior was the same whether `{}` was present before or not (both invoked the default constructor), but a real constructor is clearer.

native-activity/app/src/main/cpp/main.cpp
DanAlbert: One of us is misreading the diff, and I think it's you. This is the end of the class definition. The diff makes it look like a stray `;` at the end of a function because it's not-so-helpfully making this look like an edit to completely unrelated code.

PR #1143, opened 2025-09-17
Increase warning level in curl-ssl.

DanAlbert:


DanAlbert:
> do we have any usage numbers?

None that would be useful, if I even had access to them. We wouldn't be tell the difference between a GHA build running every ten minutes but never shipping to anyone vs an APK that builds once per OpenSSL update and ships to a billion users.

> i'm curious whether this is still the least worst/difficult way to get openssl in your android app, and we should at least try to maintain openssl even if we've given up ever having the staffing to do more packages.

It would be the least worst if we could afford to maintain it. We can't, so it isn't, because it's always out of date.

PR #1163, opened 2026-01-29
Support camera callback parameter added in r30

emrekultursay:
Without this change, build fails with NDK r30 due to missing field in initialization of the `ACameraDevice_stateCallbacks` object.


enh-google (approved):

emrekultursay:
Do you need to hit some button to unblock the presubmit?

PR #1169, opened 2026-02-19
Upgrade NDK to r29

Sharjeel-Khan:
r28c is not maintained anymore so I am upgrading the default NDK to r29.


pirama-arumuga-nainar (approved):

pirama-arumuga-nainar:
(cc: @enh-google )

